Issue #2, opened 2022-10-12
Website ScreenShot


StevenDufresne:
We need to show a screenshot of the website. 

## Current Implementation:
The current implementation uses [mShots](https://github.com/Automattic/mShots), a wp.com service that takes a screenshot of the website.

Here is the [current source code](https://github.com/WordPress/wordpress.org/blob/a125fef3538f1d8fa31b2adee76893fbbde97891/wordpress.org/public_html/wp-content/themes/pub/wporg-showcase/functions.php#L38) for handling screenshots.

**Basic Logic:**
- Check if a URL is associated with the post, use that.
- Otherwise, grab the site domain and use that.
- Output an `<img>` tag with the mShots URL.

## Problems:
- If the site doesn't have an image cached by mShots, a new one is generated but mShots displays a placeholder image. Sometimes users will see that if they are the first one to load the page and an image is not generated. When they reload, it will be replace with the screenshot.
- Sometimes screenshot have popups (usually eu cookie acceptance ones) showing up. 

## Solutions:
We have a react component that loads mShots and handles the placeholder problem. We actually have 2 right now: [wporg-mu-plugins](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/blocks/screenshot-preview), [pattern-directory](https://github.com/WordPress/pattern-directory/pull/528).

There are currently no solutions in place to fix popups in screenshots, part of this story will be to figure out what to do about that. Do we want to leave the problem with third party sites? Maybe we can instruct third parties to use querystrings and hide their popups. Idk, TBD.

PR #20, opened 2022-10-24
Add a site screenshot block.

StevenDufresne:
See #2 

This PR ports the screenshot code from the old showcase themes and block-ifies it. 

## Alternate Approaches
I looked into reusing the [screenshot-preview](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/blocks/screenshot-preview) block but decided against it right now. This code works properly and accounts for the `screenshot` meta URL & fallback better than the block currently does. Additionally, the block has an initial loading state which is not necessary here since only one image is loaded. 

That said, there is still the issue where unpreviewed URLs return the mShots placeholder (see #2 for more information). 

Since we are still waiting for design, very little design has been applied.


source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @param array    $attributes Block attributes.
+ * @param string   $content    Block default content.
+ * @param WP_Block $block      Block instance.
+ *
+ * @return string Returns the block markup.
+ */
+function render() {
+	global $post;
+	$width = 400;
```
StevenDufresne, line 40: This is bound to change later.

renintw:
This works correctly. But I got a few questions (some are probably silly..)

1. Does it really get value from `$screenshot = get_post_meta( $post->ID, 'screenshot', true );`, I don't see the meta box of screenshot in post editor. Is it basically just falling back to use the domain?
2. > Additionally, the block has an initial loading state which is not necessary here since only one image is loaded.

If we use multiple blocks in a grid on the front page, I'm thinking the loading state would be of help then?

3. > That said, there is still the issue where unpreviewed URLs return the mShots placeholder (see https://github.com/WordPress/wporg-showcase/issues/2 for more information)

This seems to need some JS or React to solve it. I'm thinking if we can follow the logic in https://github.com/WordPress/pattern-directory/pull/528 as their layout seem similar?

StevenDufresne:
> Does it really get a value from `$screenshot = get_post_meta( $post->ID, 'screenshot', true );`, I don't see the meta box of the `screenshot` in post editor. Is it basically just falling back to use the site domain?
> 

They show up as custom fields. Not all posts have them.

Screen Shot 2022-10-25 at 9 36 04 AM

> If we use multiple blocks in a grid on the front page, would the loading state be of help then?
> 

There is a lot of caching going on, jetpack actually caches the images. See `jetpack_photon_url`. So they should load immediately.

> Off the top of my head, this seems inevitable to need some JS or React to solve it. I'm thinking to follow the logic in [WordPress/pattern-directory#528](https://github.com/WordPress/pattern-directory/pull/528) as the layouts seem similar, does it make sense to do so?

Maybe, we could also run a job that makes the requests periodically, or under a certain circumstances (like saving a new post) to limit this from happening. It's pretty uncommon. If you take a look at wordpress.org/showcase right now, you will have a really hard time finding one.

> I can only come up with a hassling/inefficient method at the moment if we want to do something on our side - Before making a screenshot, `document.body.innerHTML.search("Accept Cookie");` the site and try to remove the popups..

Yeah, this is a real challenge. It probably makes sense to pull this out into it's own issue altogether.

Issue #5, opened 2022-10-12
Related Sites


StevenDufresne:
This is currently provided by JetPack. 

Link: [Block Info](https://jetpack.com/support/related-posts/)
Link: [JetPack Filters](https://jetpack.com/support/related-posts/customize-related-posts/)
Link: [Replacing fallback image](https://jetpack.com/blog/add-a-default-fallback-image-if-no-image/) 

 
## Design
Link: https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=850%3A17307

<img width="677" alt="Screen Shot 2022-10-31 at 9 57 17 AM" src="https://user-images.githubusercontent.com/1657336/198911286-7280390f-b8ee-4ebf-9bf6-e3b51131f8a6.png">

StevenDufresne:
Re-opening because of a JetPack upgrade to 11.1 removed the injected version for block themes. 

There is still an issue with the title as defined here though:
https://github.com/Automattic/jetpack/issues/27108

StevenDufresne:
The title still shows up when there are no related posts, but we shouldn't have any issues with that since posts should be categorized and tagged properly. Closing this for now.

PR #21, opened 2022-10-24
Filter jetpack images and apply mShots image as default.

StevenDufresne:
This PR makes use of the `jetpack_images_get_images` filter to provide the `fallback` image for the related post. The `fallback` image is the `mShots` URL.

This is related to #20 as it will use the same functions which are [pre-existing](https://github.com/WordPress/wordpress.org/blob/891d2d7244ff32d04092c034a40bfdac3813bc64/wordpress.org/public_html/wp-content/themes/pub/wporg-showcase/functions.php#L19) with some minor refactoring.

See: #5

## Test
You'll need to test this in a sandbox or figure out how to configure JetPack locally.


StevenDufresne:
I'm going to merge this in as another PR relies on it. A code review would still be very much appreciated.

PR #23, opened 2022-10-25
Update subnav to have the correct menu.

### LINKED ISSUE #1: Add wporg submenu
**Menu Items**
- Submit a site

**Acceptance criteria**
- Clicking on "Showcase" should go to showcase home
- If we are viewing, a tag/category/search archive it should have the appropriate word added 
  - Ie: "Showcase &middot; Results"
- The subnav should stretch full screen( [Ref](https://github.com/WordPress/wporg-main-2022/issues/129#issuecomment-1262750330) )

I'm not sure how to do this without a custom block. I looked into doing something like [local-navigation](https://github.com/WordPress/wporg-news-2021/blob/trunk/source/wp-content/themes/wporg-news-2021/block-template-parts/local-navigation.html) and maybe use [get_the_archive_title](https://developer.wordpress.org/reference/hooks/get_the_archive_title/) but it's also used on the search page which doesn't call that function. I don't see a hook for modifying the search title.

# Designs
We have a few versions:
Screen Shot 2022-10-31 at 10 28 41 AM
Screen Shot 2022-10-31 at 10 28 31 AM
Screen Shot 2022-10-31 at 10 28 20 AM
Screen Shot 2022-11-14 at 2 11 17 PM

StevenDufresne:
This PR:
- Adds build step to create a menu
- Creates a pattern for the subnav

See: #1 

## Screenshots
Screen Shot 2022-10-25 at 10 59 22 AM

## Issues
- Since the navigation block appears to need a `ref` that is environment specific, the pattern will be _broken_ upon environment creation.
- The [figma designs](https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=535%3A4737) are trending to be white, so this pr won't close #1.


StevenDufresne:
@ryelle Does it make sense to do it like this or have it spread across the page patterns like how [wporg-main-2022](https://github.com/WordPress/wporg-main-2022/blob/5809e44be957dbe4181ffab163fe902492e77c9d/source/wp-content/themes/wporg-main-2022/patterns/download-releases.php).

StevenDufresne:
> @ryelle Does it make sense to do it like this or have it spread across the page patterns like how [wporg-main-2022](https://github.com/WordPress/wporg-main-2022/blob/5809e44be957dbe4181ffab163fe902492e77c9d/source/wp-content/themes/wporg-main-2022/patterns/download-releases.php) does it.

source/wp-content/themes/wporg-showcase-2022/parts/header.html
```diff
@@ -1,2 +1,2 @@
 <!-- wp:wporg/global-header /-->
-<!-- wp:pattern {"slug":"wporg-parent-2021/subnav-bar"} /-->
+<!-- wp:pattern {"slug":"wporg-showcase-2022/subnav"} /-->
```
ryelle, line 2: What do you think about keeping this as `/subnav-bar`? Mostly as a devex thing, so we know it's meant to be a fork of the parent's pattern.

ryelle (commented):
> Does it make sense to do it like this or have it spread across the page patterns like how [wporg-main-2022](https://github.com/WordPress/wporg-main-2022/blob/5809e44be957dbe4181ffab163fe902492e77c9d/source/wp-content/themes/wporg-main-2022/patterns/download-releases.php).

If these pages end up edited in the editor like the Download pages, the pattern "blocks" will be replaced with the pattern content, and when they're synced, `download-releases` is what happens.

I mentioned this issue in [the followup to pattern-dir blockification](https://github.com/WordPress/pattern-directory/pull/518#issuecomment-1265789522), when I tried to get the syncing to work. If you're not worried about syncing yet, this should be fine. I'm hoping we can figure out a way to make this approach work since it _would_ be nicer to keep repeated-use content in reusable patterns/templates.

source/wp-content/themes/wporg-showcase-2022/parts/header.html
StevenDufresne: Yep. Fine by me.

StevenDufresne:
> I mentioned this issue in [the followup to pattern-dir blockification](https://github.com/WordPress/pattern-directory/pull/518#issuecomment-1265789522), when I tried to get the syncing to work. If you're not worried about syncing yet, this should be fine. I'm hoping we can figure out a way to make this approach work since it _would_ be nicer to keep repeated-use content in reusable patterns/templates.

Yeah, I find it tricky to insert the pattern, then export it to my own pattern so it can be translated with _my text_. Should my local one be allowed to be inserted? Or should only the parent one? 

To be honest, the pattern import/export sequences are a bit of a brainbuster for me. They make better sense if all development happens in the site editor (minus the translated workarounds), but I guess since I still have a mental model of building templates in my code editor with the help of the site editor I'm tripping myself up... just thinking out loud here. :)

Issue #9, opened 2022-10-19
Search Bar


StevenDufresne:
We need a search bar component that redirects users to the search/archive page. 

## Designs

<img width="443" alt="Screen Shot 2022-10-31 at 10 24 24 AM" src="https://user-images.githubusercontent.com/1657336/198913179-1f915a0e-d4c8-4cb8-909c-01cfd6b1c7db.png">

StevenDufresne:
Still need to address redundancy with labels noted in #25.

StevenDufresne:
Comment added to designs: 
https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA?node-id=1256:21493#301126354

PR #25, opened 2022-10-26
Add Searchbar Pattern/Block Styles

StevenDufresne:
This PR adds a pattern for the search bar which extends `core/search` with block styles. This PR also introduces a style component that imports a style sheet the same way we do in `wporg-main-2022`. See [this pull request](https://github.com/WordPress/wporg-main-2022/pull/125) for background info.

To be honest, this is the first time I have created block styles and I'm okay if there's a better `theme.json` way of doing it and I have to backtrack a bit. I had to use a lot of `!important` declarations to fight specificity. I'm wondering if that's the result of this being the wrong approach.

Additionally, I called this the secondary search bar because based on history, there is usually a large one that is included in the header. This search bar style seems to be only used in archive-like scenarios.

See: #9

## Variations

Note: Disregard the placeholder "Search Patterns". Updated the placeholder after the screenshots. 😆 

**Label, Button Inside, Icon** _This variation is currently in the designs._
Screen Shot 2022-10-26 at 3 14 48 PM

**Label, Button Inside, Text** 
Screen Shot 2022-10-26 at 3 14 36 PM

**Label, Button Outside, Icon** 
Screen Shot 2022-10-26 at 3 14 01 PM

**Label, Button Outside, Text** 
Screen Shot 2022-10-26 at 3 14 09 PM

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/patterns/secondary-search-bar.php
+++ b/source/wp-content/themes/wporg-showcase-2022/patterns/secondary-search-bar.php
@@ -7,4 +7,4 @@
 
 ?>
 
-<!-- wp:search {"label":"Search","showLabel":true,"placeholder":"<?php esc_attr_e( 'Search sites...', 'wporg' ); ?>","width":100,"widthUnit":"%","buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true,"className":"is-style-secondary-search-control"} /-->
+<!-- wp:search {"label":"<?php esc_attr_e( 'Search', 'wporg' ); ?>","showLabel":true,"placeholder":"<?php esc_attr_e( 'Search sites...', 'wporg' ); ?>","width":100,"widthUnit":"%","buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true,"className":"is-style-secondary-search-control"} /-->
--- a/source/wp-content/themes/wporg-showcase-2022/patterns/secondary-search-bar.php
+++ b/source/wp-content/themes/wporg-showcase-2022/patterns/secondary-search-bar.php
@@ -1,0 +1,10 @@
+<?php
+/**
+ * Title: Subnav
+ * Slug: wporg-showcase-2022/secondary-search-bar
+ * Categories: wporg
+ */
+
+?>
+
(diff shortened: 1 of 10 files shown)


ryelle:
> To be honest, this is the first time I have created block styles and I'm okay if there's a better theme.json way of doing it and I have to backtrack a bit. I had to use a lot of !important declarations to fight specificity. I'm wondering if that's the result of this being the wrong approach.

Unfortunately there isn't a theme.json-y way of working with block styles, so what you've got is probably the best even with all the `important`s.

A scan of the code looks good, but I don't have a local env for showcase right now, so I'll leave the detailed review for someone else :)

adamwoodnz:
Lacking experience here but yeah the swath of `!important`s certainly seems like a code smell. Sounds like it can't be helped but maybe something worth raising with Gutenberg folks?

From an a11y standpoint might be worth reviewing whether placeholder is necessary. Some recent feedback on Learn suggested it was too verbose: https://wordpress.slack.com/archives/C02RP4X03/p1663279430556099

Same as @ryelle with the local env, maybe @renintw can run it?

StevenDufresne:
>From an a11y standpoint might be worth reviewing whether placeholder is necessary. Some recent feedback on Learn suggested it was too verbose: https://wordpress.slack.com/archives/C02RP4X03/p1663279430556099

The latest designs remove the label so I'll turn it off.

source/wp-content/themes/wporg-showcase-2022/patterns/secondary-search-bar.php
```diff
@@ -0,0 +1,10 @@
+<?php
+/**
+ * Title: Secondary Search Bar
+ * Slug: wporg-showcase-2022/secondary-search-bar
+ * Categories: wporg
+ */
+
+?>
+
+<!-- wp:search {"showLabel":false,"placeholder":"<?php esc_attr_e( 'Search sites...', 'wporg' ); ?>","width":100,"widthUnit":"%","buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true,"className":"is-style-secondary-search-control"} /-->
```
ryelle, line 10: `"showLabel":false` hides the label visually, but it's still there in the content (using the default label, "Search"), so the feedback about [the overly verbose placeholder from here](https://wordpress.slack.com/archives/C02RP4X03/p1663279430556099) still applies.

StevenDufresne: I should have verified. Thanks.

StevenDufresne:
Opened issue specific to redundant voice over:
https://github.com/WordPress/wporg-showcase-2022/issues/39

Issue #18, opened 2022-10-21
Archive/Search context sentence.


StevenDufresne:
Archive and search pages have a sentence which contextualized the results.

# Designs
Link: https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=878%3A11413

**Search**
<img width="148" alt="Screen Shot 2022-11-02 at 3 50 10 PM" src="https://user-images.githubusercontent.com/1657336/199420096-352fab28-ccc9-4ab5-98f5-7beb257510d6.png">

**Tags**
<img width="305" alt="Screen Shot 2022-11-02 at 3 50 04 PM" src="https://user-images.githubusercontent.com/1657336/199420101-62768135-6250-49c1-9b88-d8051c09e7dc.png">

Issue #19, opened 2022-10-21
Layout: Search


StevenDufresne:
Contains:
- #1 
- #18 
- #9 

# Designs
Link: https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=945%3A17044

<img src="https://user-images.githubusercontent.com/1657336/200215147-f789b511-04c0-4930-b9c5-80c25e6fe1a5.png" width="400px" />

Issue #14, opened 2022-10-19
Layout: Archive


StevenDufresne:
Contains:
- #1 
- #18 
- #9 
- #12 

# Designs
![image](https://user-images.githubusercontent.com/1657336/198908512-aba390c5-055c-47de-93e0-204d379f1feb.png)

**Mobile**
<img src="https://user-images.githubusercontent.com/1657336/198908578-a00a1429-6185-4682-8b0d-75bcd230234b.png" width="300px">

PR #32, opened 2022-11-04
Add a context component for Archive and search.

StevenDufresne:
This PR does 2 things:
- Creates a component as defined in #18 
- Creates a pattern where the component is grouped with search bar
  - Adds relevant CSS

**Notes**
I would have like to not add custom CSS, but it was not possible to do through the editor. 

Fixes: #18 
Related: #14, #19

## Tags
| Mobile | Desktop |
| --- | --- |
| Screen Shot 2022-11-04 at 12 49 34 PM | Screen Shot 2022-11-04 at 12 49 50 PM | 

## Search
| Mobile | Desktop |
| --- | --- |
| Screen Shot 2022-11-04 at 12 49 00 PM | Screen Shot 2022-11-04 at 12 52 50 PM | 

## Archive
| Mobile | Desktop |
| --- | --- |
| Screen Shot 2022-11-04 at 12 54 17 PM | Screen Shot 2022-11-04 at 12 53 53 PM |


source/wp-content/themes/wporg-showcase-2022/templates/search.html
```diff
@@ -1,37 +1,37 @@
 <!-- wp:template-part {"slug":"header" } /-->
 
 <!-- wp:group {"tagName":"main","style":{"spacing":{"blockGap":"0px"}},"className":"entry-content","layout":{"type":"constrained"}} -->
-<main class="wp-block-group entry-content">
-	<!-- wp:query-title {"type":"search"} /-->
-	<!-- wp:pattern {"slug":"wporg-showcase-2022/secondary-search-bar"} /-->
+<main class="wp-block-group entry-content"><!-- wp:group {"align":"wide","layout":{"type":"default"}} -->
+	<div class="wp-block-group alignwide">
+		<!-- wp:pattern {"slug":"wporg-showcase-2022/results-bar"} /-->
```
StevenDufresne, line 6: This is the only change that really matters.

The other layout change help the template with the layout but will ultimately be addressed later.

source/wp-content/themes/wporg-showcase-2022/src/archive-results-context/index.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Internal dependencies
+ */
+import metadata from './block.json';
+
+function Edit( { attributes, name } ) {
+	return (
+		<div { ...useBlockProps() }>
+			<Disabled>
```
ryelle, line 17: You probably don't need the `Disabled` wrapper, unless there will be links in the text later?

source/wp-content/themes/wporg-showcase-2022/src/archive-results-context/index.php
```diff
@@ -0,0 +1,75 @@
...
+				'<p>There are <b>%1$s</b> sites tagged with <b>%2$s</b></p>',
+				$wp_query->found_posts,
+				'wporg'
+			),
+			number_format_i18n( $wp_query->found_posts ),
+			esc_html( get_queried_object()->name )
+		);
+	}
+
+	return sprintf(
```
ryelle, line 49: I'd recommend doing the same updates here that I added in https://github.com/WordPress/wporg-showcase-2022/pull/33 to support the various block `supports` properties, it will make working with this block in templates easier in the future.

source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
```diff
@@ -127,3 +127,21 @@ button[type="submit"] {
 	}
 
 }
+
+@media (max-width: 600px) {
```
ryelle, line 131: Why add this here rather than a `style.scss` in the block?

edit: I just re-read the description, and see that you'd rather not use CSS… can you use a columns block instead? that should switch to stacked at 600px. It just wouldn't keep the custom order.

source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
```diff
@@ -127,3 +127,21 @@ button[type="submit"] {
...
+		width: 100%;
+	}
+
+	.wporg-results-bar-container > *:first-child {
+		order: 2;
+	}
+
+	.wporg-results-bar-container > *:last-child {
+		order: 1;
+	}
```
ryelle, line 146: Changing order like this breaks the expected pattern for keyboard navigation, but any other ideas to fix it and keep the visual display are pretty hacky… Not a blocker, but not great 🤷🏻

ryelle (commented):
I got search & tags to work, but the category page still used the parent theme, I think. Also, not sure how to get to just the "archive" page, since the front page is the single-column list of posts too.

But I don't think that's a fault of this PR, just GB templates 🤷🏻

source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
StevenDufresne: Can you be more specific on how? Aren't the form controls the only focusable elements? Noting, the order will only be changed on mobile.

source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
StevenDufresne: These styles are specific to the combination of the block and the search bar. The block itself is just the archive context sentence.

StevenDufresne:
> I got search & tags to work, but the category page still used the parent theme, I think. Also, not sure how to get to just the "archive" page, since the front page is the single-column list of posts too.
> 
> But I don't think that's a fault of this PR, just GB templates 🤷🏻

Yeah, because the parent has a `category.html` template, we'll need to provide one to override it. I think I'll just move the archive content into a pattern and load that in all the "archive" like pages: tags, category, archive, search.

source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
ryelle: Fair question, I wasn't clear - I said keyboard nav but was really talking about how the screen reader focus moves. A screen reader will read in source-order, so in this case the context bar will be read first, then it'll skip back to the search, then skip the context to get to the content. And that's the case for mobile screen readers too.

source/wp-content/themes/wporg-showcase-2022/src/archive-results-context/index.js
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Internal dependencies
+ */
+import metadata from './block.json';
+
+function Edit( { attributes, name } ) {
+	return (
+		<div { ...useBlockProps() }>
+			<ServerSideRender block={ name } attributes={ attributes } />
+		</div>
```
ryelle, line 17: It seems like this wrapper in JS is not needed, since the rendered block also contains the block props attributes.

I added this to a page to test out the attributes, and noticed my "3X-Small" padding is there twice.
Screenshot 2022-11-07 at 3 39 53 PM

```suggestion
		<ServerSideRender block={ name } attributes={ attributes } />
```

ryelle:
I tried the columns approach [I mentioned here](https://github.com/WordPress/wporg-showcase-2022/pull/32#discussion_r1014487387), which removes the need for the CSS (but doesn't change the order).

Not adding directly to the PR in case you didn't go this way for a reason :)

```html
<!-- wp:columns {"verticalAlignment":"center","style":{"spacing":{"padding":{"top":"var:preset|spacing|40","bottom":"var:preset|spacing|40"}}}} -->
<div class="wp-block-columns are-vertically-aligned-center" style="padding-top:var(--wp--preset--spacing--40);padding-bottom:var(--wp--preset--spacing--40)"><!-- wp:column {"verticalAlignment":"center","width":"50%"} -->
<div class="wp-block-column is-vertically-aligned-center" style="flex-basis:50%"><!-- wp:wporg/archive-results-context /--></div>
<!-- /wp:column -->

<!-- wp:column {"verticalAlignment":"center","width":"50%"} -->
<div class="wp-block-column is-vertically-aligned-center" style="flex-basis:50%"><!-- wp:search {"showLabel":false,"placeholder":"<?php esc_attr_e( 'Search sites...', 'wporg' ); ?>","width":100,"widthUnit":"%","buttonText":"<?php esc_attr_e( 'Search', 'wporg' ); ?>","buttonPosition":"button-inside","buttonUseIcon":true,"className":"is-style-secondary-search-control"} /--></div>
<!-- /wp:column --></div>
<!-- /wp:columns -->
```

StevenDufresne:
> ```
> 
> <div class="wp-block-columns are-vertically-aligned-center" style="padding-top:var(--wp--preset--spacing--40);padding-bottom:var(--wp--preset--spacing--40)">
> <div class="wp-block-column is-vertically-aligned-center" style="flex-basis:50%"></div>
> 
> 
> 
> <div class="wp-block-column is-vertically-aligned-center" style="flex-basis:50%"></div>
> </div>
> 
> ```

Yeah, I didn't go this route solely because of the ordering change seen in the designs.

source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
StevenDufresne: I think we'll going to need this container functionality for other

> ```
> 
> <div class="wp-block-columns are-vertically-aligned-center" style="padding-top:var(--wp--preset--spacing--40);padding-bottom:var(--wp--preset--spacing--40)">
> <div class="wp-block-column is-vertically-aligned-center" style="flex-basis:50%"></div>
> 
> 
> 
> <div class="wp-block-column is-vertically-aligned-center" style="flex-basis:50%"></div>
> </div>
> 
> ```

Additionally, the one problem I see with this approach is that the search bar will be variable width after passing the mobile breakpoint. We can control its width with column percentages but since they are percentage based, it will be a different sized component at 600px and 900px.  We also can't give it a width because it needs to be `100%` on mobile.

ryelle: You can set the column width in other units, like px or rem, and remove the width from the other block which will default it to the rest of the space. On smaller screens, it will still expand to the 100% width. Does that help?

source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
StevenDufresne: It does! Well done.

I'm going to merge this now.

PR #30, opened 2022-11-03
Add a block style to control edge-spacing.

StevenDufresne:
I'm having a difficult time creating consistency between the edges of the browser and containers. This PR adds a block style for groups that will allow us to apply the `edge-spacing` CSS to ensure consistency across components and pages. I _think_ this makes sense.

The calculation used for the edge spacing was taken from `wporg-news-2021`:
```
clamp( 24px, calc( 100vw / 18 ), 80px )
```

# Screenshot

| Mobile | Desktop |
| --- | --- |
| Screen Shot 2022-11-03 at 11 53 07 AM |  Screen Shot 2022-11-03 at 12 02 41 PM |

PR #34, opened 2022-11-04
Try adding inline space to all blocks when on smaller screens

ryelle:
Alternative approach to #30 — Since this behavior (space around the blocks) should probably be universal across pages, I've updated the media query to add a margin to blocks when the page size is about to run into the sides. I think this might be a better approach to the issue, rather than needing to apply the block style on each page.

I also updated the single template slightly to remove an unnecessary group, and the page template to add the layout properties to `wp:post-content` — using this method also adds support for wide & full alignment in post content.

For example, I added a wide-width box to the showcase site template. It stays wide until it hits the edges, then it stays edge-to-edge all the way to mobile screens.

(I only took 2 page screenshots, there was less obvious difference there)

| Size | Showcase site | Page |
|------|---------------|------|
| 1400 |  | - |
| 800 |  |  |
| 600 |  | - |
| 400 |  |  |

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
+++ b/source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
@@ -111,19 +111,12 @@ button[type="submit"] {
 	}
 }
 
-@media (max-width: 1199px) {
-
-	/* Once we start touching content, add in some side padding. */
-	.entry-content.entry-content > .alignwide,
-	.entry-content.entry-content > .alignfull {
-		padding-left: var(--wp--preset--spacing--60) !important;
-		padding-right: var(--wp--preset--spacing--60) !important;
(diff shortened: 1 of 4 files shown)


StevenDufresne:
What would be the block hierarchy for a component like this:

Screen Shot 2022-11-07 at 9 16 18 AM

StevenDufresne:
> What would be the block hierarchy for a component like this:


Tried this out. I guess we would apply `left` and `right` padding of "large" and it wouldn't use this CSS because we can't use `is-layout-constrained`. I'm ok with that but it's not completely self-documenting though. 

I think overall I'm ok with this since block styles probably shouldn't be coerced into doing utility-like CSS as I have it in #30. You also can't combine block styling so it would probably lead to more `core/groups` if we did use a block style.

👍

Issue #29, opened 2022-11-01
Single: Site details box


StevenDufresne:
I'm not certain we have agreed on which fields will exist here.

# Designs
Link: https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=850%3A17307

<img width="409" alt="Screen Shot 2022-11-01 at 3 29 54 PM" src="https://user-images.githubusercontent.com/1657336/199172844-e1bfc956-c85b-4184-8f86-6116621fa9f3.png">

Issue #4, opened 2022-10-12
Layout: Single


StevenDufresne:
The single template will be used to display information about a specific website.

**Notable Features:**
- Displays post content
- Displays associated tags
- Includes https://github.com/WordPress/wporg-showcase/issues/5
- Includes https://github.com/WordPress/wporg-showcase/issues/6
- Includes https://github.com/WordPress/wporg-showcase-2022/issues/29

## Design
Link: https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=1256%3A21116

<img src="https://user-images.githubusercontent.com/1657336/200217682-e4156a78-2282-4214-948c-9384ac9fd14c.png" width="400px">

PR #35, opened 2022-11-07
Try shortcode for information box.

StevenDufresne:
This PR adds the metadata box to the single template.

Fixes: #29 
Related #4.

## Notes
- I separate the items into 2 `div`s.
  - I tried to use `grid` or `flex` displays but I would also run into issues getting spacing right and having the right control over the order, etc.

# ScreenShots
| Mobile | Desktop |
| --- | --- |
| Screen Shot 2022-11-10 at 3 37 53 PM | Screen Shot 2022-11-10 at 3 21 34 PM | 

## Site with no meta data
Screen Shot 2022-11-10 at 3 24 39 PM


StevenDufresne:
@jasmussen @javierarce 

The original designs appear to link the metadata, presumable to archives. I don't think there is a compelling reason why we should allow those meta fields to be archives. Most of them will be sparse. Any opinion about how it looks without them being links? [See screenshots in description].

source/wp-content/themes/wporg-showcase-2022/src/site-meta-list/index.php
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Get's tags and categories associated with site.
+ *
+ * @param integer $post_id
+ * @return string
+ */
+function get_associated_terms( $post_id ) {
+	$terms = array();
+	$tags = get_the_terms( $post_id, 'post_tag' );
+	$categories = get_the_terms( $post_id, 'category' );
```
StevenDufresne, line 103: I feel like there should be a better wp way to get all the tags and categories.

source/wp-content/themes/wporg-showcase-2022/templates/single.html
```diff
@@ -11,6 +11,13 @@
 <p style="margin-top:var(--wp--preset--spacing--20)"><a href="[domain]">[pretty_domain]</a></p>
 <!-- /wp:paragraph -->
 
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|20","right":"var:preset|spacing|20","bottom":"var:preset|spacing|20","left":"var:preset|spacing|20"},"margin":{"top":"var:preset|spacing|30"}}},"backgroundColor":"blueberry-4"} -->
+<div class="wp-block-group has-blueberry-4-background-color has-background" style="margin-top:var(--wp--preset--spacing--30);padding-top:var(--wp--preset--spacing--20);padding-right:var(--wp--preset--spacing--20);padding-bottom:var(--wp--preset--spacing--20);padding-left:var(--wp--preset--spacing--20)"><!-- wp:heading {"style":{"typography":{"fontStyle":"normal","fontWeight":"600"}},"fontSize":"normal","fontFamily":"inter"} -->
+<h2 class="has-inter-font-family has-normal-font-size" style="font-style:normal;font-weight:600">More about this site</h2>
```
StevenDufresne, line 16: These will probably need to be moved into a pattern for translating. Will do it in a follow up pr.

jasmussen:
>  Any opinion about how it looks without them being links? [See screenshots in description].

Looks fine to me. I suppose we can always revisit, right?

source/wp-content/themes/wporg-showcase-2022/src/site-meta-list/index.php
bengreeley: @StevenDufresne I wonder if you could use `wp_term_query()` which accepts an array of taxonomies? https://developer.wordpress.org/reference/classes/wp_term_query/

StevenDufresne: It doesn't appear like you can filter them based on the current post. I would probably have to recreate some of the `get_the_terms` logic:

tellyworth: I think you could use `wp_get_object_terms()` to do that, since it optionally takes an array of `$taxonomy` names.

`get_the_terms()` shows how to use `wp_get_object_terms()`, you could start with the guts of that. In fact if it wasn't for its buggy cache call, `get_the_terms( $post_id, array('post_tag', 'category) )` would probably work.

I'm not sure if there's a performance advantage to doing it that way though. Your way is easy to read.

dd32: `wp_get_object_terms()` can return Terms from multiple Taxonomies, but it bypasses the caching that `get_the_terms()` has. Realistically I think it's rare that terms from multiple taxonomies would be output in the same data..

source/wp-content/themes/wporg-showcase-2022/src/site-meta-list/index.php
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	if ( empty( $terms ) ) {
+		return '';
+	}
+
+	$links  = array();
+
+	foreach ( $terms as $value ) {
+		$links[] = "<a href='" . get_term_link( $value->term_id, $value->taxonomy ) . "'>" . esc_html( $value->name ) . '</a>';
```
dd32, line 118: Term names are intended to be saved as "safe". This might be problematic if there's ever a term with `'` or other non-ascii symbols.

`get_term_link()` can theoretically return a WP_Error, but I don't think you need to account for that. Some would say you need to `esc_url()` the URL, but I don't believe you need to.

See https://developer.wordpress.org/reference/functions/get_the_term_list/ for some logic to mimmick.

```suggestion
		$links[] = "<a href='" . get_term_link( $value->term_id, $value->taxonomy ) . "'>" . $value->name . '</a>';
```

source/wp-content/themes/wporg-showcase-2022/src/site-meta-list/index.php
```diff
@@ -0,0 +1,138 @@
...
+		return '';
+	}
+
+	$list_items = array();
+
+	$meta_fields = array(
+		array(
+			'label' => __( 'Author', 'wporg' ),
+			'key' => 'author',
+		),
```
dd32, line 33: It would be nice if this, and all the other code was using consistent indentation of array items. Same for variable alignments.
ie.
```suggestion
		array(
			'label' => __( 'Author', 'wporg' ),
			'key'   => 'author',
		),
```

renintw (commented):
Code and Layout were tested and look good 👍 
One thing: do you think we could use `grid` like [this](https://github.com/WordPress/wporg-showcase-2022/compare/try/site-details-info...test)? 
It looks a bit more straightforward to me.
Here's the testing result:
| Orignal | New |
| -------| ------|
| Screen Shot 2022-11-16 at 12 46 43 PM | Screen Shot 2022-11-16 at 12 46 47 PM |

source/wp-content/themes/wporg-showcase-2022/src/site-meta-list/index.php
StevenDufresne: Right. I just let our [wporg-repo-tools](https://github.com/WordPress/wporg-repo-tools/blob/trunk/configs/phpcs.xml.dist) configuration control that. 

Running `composer run format [...]` didn't care about indentation.

StevenDufresne:
> Code and Layout were tested and look good 👍 One thing: do you think we could use `grid` like [this](https://github.com/WordPress/wporg-showcase-2022/compare/try/site-details-info...test)? It looks a bit more straightforward to me. Here's the testing result:

Tag-wise, I think a `<dl>` list is semantically more precise?? I think ultimately the examples are very similar. You'll need to add margin into your example to control layout on mobile where it's stacking. This PR adds it to the `dd` because it already has some default browser margins that we need to nullify.

StevenDufresne:
@renintw Are there any changes that you would like to see? Do you want to update the grid?

renintw:
> @renintw Are there any changes that you would like to see? Do you want to update the grid?

No, I think it makes sense to keep dl dd for semantic purpose as well. I'm ok with what it is right now.

renintw (approved):

PR #38, opened 2022-11-09
Reorganize the archive templates.

StevenDufresne:
Fixes: #14 

This PR creates a pattern to be used in the archive templates:
- category.html
  - Needed because the parent has one 
- archive.html
- index.html
  - For some reason, this is called for the Posts page. I don't know why it doesn't use archive.html?
- search.html

| Mobile | Tablet | Desktop |
| --- | --- | --- |
| Screen Shot 2022-11-10 at 11 25 19 AM | Screen Shot 2022-11-10 at 11 25 35 AM |Screen Shot 2022-11-10 at 11 25 50 AM |

 Designs: https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=1256%3A21493


renintw (approved):
Tested and looks good 👍

StevenDufresne:
> > For some reason, this is called for the Posts page. I don't know why it doesn't use archive.html?
> 
> I'm thinking it would go this way image

Oh that's right. Thanks!

PR #41, opened 2022-11-14
Create a template for the page banners.

### LINKED ISSUE #10: Layout: Submit Site Assets
We have decided to hold off on launching this page. See #41 for context.

<hr>

This page contains assets that showcase site owners can download.

Current page: https://wordpress.org/showcase/submit-a-wordpress-site/graphics/.

# Designs

image

### Mobile
Link: https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=884%3A12798

StevenDufresne:
We have decided to keep this open assuming this isn't needed.

<hr>

This PR adds `page-banners.html` and `page-banners.php`.

The implementation slightly diverges from the designs as the designs use a larger content width than we currently have established in the parent theme.

This doesn't include the fix for the "Showcase · Banners".

Fixes: #10

| Mobile | Tablet | Desktop |
| --- | --- | --- |
| Screen Shot 2022-11-14 at 3 11 02 PM  | Screen Shot 2022-11-14 at 3 11 14 PM | Screen Shot 2022-11-14 at 3 11 33 PM | 

**Designs**
image


outdoor2kode (commented):
Tested and looks the same as the PR description on mobile/tablet/desktop 👍 
Got two questions:

Do you think it makes sense to add `width: 781px` to the text `If your site is...` and change from `group block` to `stack block` which makes it able to justify items to the center?
image
image

Curious if we could use `.svg` as the image source? The img looks a bit blurry to me on a high-resolution screen.
image

StevenDufresne:
> Tested and looks the same as the PR description on mobile/tablet/desktop 👍 Got two questions:
> 
> Do you think it makes sense to add `width: 781px` to the text `If your site is...` and change from `group block` to `stack block` which makes it able to justify items to the center? image image

Good idea. I have added more groups and forced the layout to `781px`.  

We don't need to change to a stack and justify center. The offset was due to the images being different widths. I've updated their widths and will look for better images to push to the cdn.

StevenDufresne:
> Curious if we could use `.svg` as the image source? The img looks a bit blurry to me on a high-resolution screen. 

Hmm, these images are meant to be downloaded by end users. I think this is maybe a discrepancy in the experience, we should probably display the banners but then provide a link to `png`s. If we did that, we could swap these `png`s out for svgs for display purposes. Ah... maybe these banners are meant to be links and clicking on them brings you to the image making it easier for you to download?

@jasmussen @javierarce What do ya'll think? Should we have separate links under the banners, [how it currently is](https://wordpress.org/showcase/submit-a-wordpress-site/graphics/), or these banners should be links? If they are to be links, I think we'll need to provide more affordance.

jasmussen:
I think if someone wants to actually include these images in their site, they can open in new tab and save as, though I wouldn't be opposed to any help text.

However I do want to resurface the question of whether we need these or not. I'm not sure including banners like these on a webpage is something people do anymore 🤔

StevenDufresne:
> However I do want to resurface the question of whether we need these or not. I'm not sure including banners like these on a webpage is something people do anymore 🤔

I don't believe I have seen that conversation. I personally don't think it's necessary.

jasmussen:
Conversation likely happened in Figma files somewhere, sorry.

StevenDufresne:
> Conversation likely happened in Figma files somewhere, sorry.

Who is in a position to make that decision? I'm totally fine with not shipping with this feature.

jasmussen:
I think we can decide here together. Another option is to start without these, then add them as _ready designs on hand_ if the need or desire reemerges.

PR #47, opened 2022-11-17
Add the ability to make the screenshot block clickable in templates.

### LINKED ISSUE #46: Images in grid should be clickable.
When we present multiple sites in a grid format, clicking on the image should navigate to the site page. They currently do not.

StevenDufresne:
Fixes #46 

This PR adds the ability to make the screenshot block clickable. This pr allows developers the ability to add `"isLink":true` to the block in the template which will wrap the image in a link and use the permalink of the associated post as the href.

This PR doesn't introduce controls in the block editor to toggle it. I'm not sure it's necessary at this point.

# Generated HTML

```
<div class="wp-block-wporg-site-screenshot">
    <a href="http://localhost:8888/post-name/">
      <img {....}>
      <span class="screen-reader-text">Post Name</span>
   </a>
</div>
```

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
@@ -53,7 +53,7 @@ function render( $attributes, $content, $block ) {
 
 	$img_content = "<img src='{$screenshot}' srcset='$srcset 2x' alt='" . the_title_attribute( array( 'echo' => false ) ) . "' />";
 
-	if ( isset( $attributes['isLink'] ) ) {
+	if ( isset( $attributes['isLink'] ) && true == $attributes['isLink'] ) {
 		$img_content .= '<span class="screen-reader-text">' . the_title_attribute( array( 'echo' => false ) ) . '</span>';
 		$img_content = '<a href="' . get_permalink( $post ) . '">' . $img_content . '</a>';
 	}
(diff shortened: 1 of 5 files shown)


source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
```diff
@@ -53,6 +53,11 @@ function render( $attributes, $content, $block ) {
 
 	$img_content = "<img src='{$screenshot}' srcset='$srcset 2x' alt='" . the_title_attribute( array( 'echo' => false ) ) . "' />";
 
+	if ( isset( $attributes['isLink'] ) && true == $attributes['isLink'] ) {
+		$img_content .= '<span class="screen-reader-text">' . the_title_attribute( array( 'echo' => false ) ) . '</span>';
```
ryelle, line 57: Since the image already has this text as the alt attribute, this creates a duplicate for the link's text, ex "Childrens Radio Foundation Childrens Radio Foundation"

ryelle (changes requested):
The attribute needs to be added to `block.json`. I would also suggest adding the toggle to the editor UI now so you don't need to come back and do it later. It doesn't need to be complicated, just a [ToggleControl](https://developer.wordpress.org/block-editor/reference-guides/components/toggle-control/).

source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
StevenDufresne: I'll remove it.

ryelle (approved):
Looks good now 👍🏻

PR #45, opened 2022-11-15
wporg submenu

renintw:
Fixes #1 

This PR adds a new block `wporg/site-breadcrumbs` to display breadcrumbs for the site.

#### Archives
| Desktop | Tablet | Mobile |
| -------- | ------ | ------- |
|  Screen Shot 2022-11-18 at 5 14 53 AM | Screen Shot 2022-11-18 at 5 15 16 AM | Screen Shot 2022-11-18 at 5 15 27 AM |

#### Search page
| Desktop | Tablet | Mobile |
| -------- | ------ | ------- |
| Screen Shot 2022-11-18 at 5 12 31 AM | Screen Shot 2022-11-18 at 5 12 48 AM | Screen Shot 2022-11-18 at 5 12 55 AM |

#### Tags
| Desktop | Tablet | Mobile |
| -------- | ------ | ------- |
| Screen Shot 2022-11-18 at 5 13 12 AM | Screen Shot 2022-11-18 at 5 13 20 AM | Screen Shot 2022-11-18 at 5 13 30 AM |

#### Categories
| Desktop | Tablet | Mobile |
| -------- | ------ | ------- |
| Screen Shot 2022-11-18 at 5 13 44 AM | Screen Shot 2022-11-18 at 5 13 52 AM | Screen Shot 2022-11-18 at 5 13 59 AM |

#### Site
| Desktop | Tablet | Mobile |
| -------- | ------ | ------- |
| Screen Shot 2022-11-18 at 5 14 16 AM | Screen Shot 2022-11-18 at 5 14 32 AM | Screen Shot 2022-11-18 at 5 14 40 AM |

PR #48, opened 2022-11-17
Apply style updates from Subpage Header Bar pattern

ryelle:
This applies the subnav style changes from the Parent theme, https://github.com/WordPress/wporg-parent-2021/commit/40f4a3dba3e477502ca7fbe5a82f986828eb239d (see https://github.com/WordPress/wporg-parent-2021/issues/55, https://github.com/WordPress/wporg-parent-2021/issues/56). The font should be `small`, the gap between items has been updated, and the current page is highlighted using a bold style.

Additionally, I've deleted the `category.html` template in the parent theme, since we've had to work around it in each child theme by duplicating the archive template. That means we can also delete that file here, and the category archive uses the child theme's `archive.html` template.

Finally, I've updated how the sticky subnav header works — I'm _assuming_ this is meant to be sticky here, since that's the behavior on the main site. There are two changes from the parent theme for this:
- Add an `is-sticky` class which is used on the Brush Stroke styled Group block, and that allows the block to be sticky.
- Add a `has-display-contents` helper class which is used on the template part block to render the template block wrapper inert style-wise — without this, the `position: sticky` on the subnav doesn't work.

### Screenshots

| Page | Screenshot |
|----|-----|
| Home (no change) | home |
| Single site, scrolled down, subnav is also sticky now | single |
| A category archive, inherits the child's `archive.html` properly | archive |
| Single page, page in nav is bold | page |

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/parts/header.html
+++ b/source/wp-content/themes/wporg-showcase-2022/parts/header.html
@@ -1,4 +1,3 @@
-<!-- wp:group {"style":{"border":{"bottom":{"color":"var:preset|color|light-grey-1","width":"1px"}}},"layout":{"type":"constrained"}} -->
-<div class="wp-block-group" style="border-bottom-color:var(--wp--preset--color--light-grey-1);border-bottom-width:1px"><!-- wp:wporg/global-header {"style":"black-on-white"} /--></div>
-<!-- /wp:group -->
+<!-- wp:wporg/global-header {"style":"black-on-white"} /-->
+
 <!-- wp:pattern {"slug":"wporg-showcase-2022/subnav-bar"} /-->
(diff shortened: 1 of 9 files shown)


StevenDufresne:
Thanks for this! Yeah looks good. 

The design team has also asked for the menu to align wide like on news.

Do you think we can do that here?

**References**
https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA?node-id=850:11695#294656240
https://github.com/WordPress/wporg-main-2022/issues/129#issuecomment-1262750330

StevenDufresne (approved):

ryelle:
> The design team has also asked for the menu to align wide like on news.

Yeah, I needed to do some updates to the header first (just PR'd that in https://github.com/WordPress/wporg-mu-plugins/pull/302). I've just pushed the parent theme & an update to the theme here to get that alignment in place.

StevenDufresne:
Do you mind also taking a look at #45 when you get a chance? It moves some of this code into a component to get breadcrumbs to work. I'm not sure what would be the best approach.

PR #49, opened 2022-11-18
Site Submission Confirmation

### LINKED ISSUE #16: Submit site confirmation
After a user submits a site, we need to display a confirmation.

# Designs
image

StevenDufresne:
Fixes #16, Related: #44

This PR adds a pattern and a template to handle site submission confirmations.

| Mobile | Tablet | Desktop |
| --- | --- | --- |
| localhost_8888_submit-a-wordpress-site_submit-site-confirmation_(iPhone 12 Pro) (1) | localhost_8888_submit-a-wordpress-site_submit-site-confirmation_(iPad Air) (1) | localhost_8888_submit-a-wordpress-site_submit-site-confirmation_ (2) |

# How To test
Note: the pattern is linking to an image in the CDN but I have yet to add it. Will do so after this is merged. You can use the image in `images` folder for local testing.
1. Create a page, set the page to use this template.
2. Visit this template via site editor and replace the missing cover image.
3. Visit your newly created page.


renintw:
Tested and looks good 👍 

One thing: Can we somehow make the text `email` be on the third line on mobile devices?  
image
Maybe add `margin: 0` for `<p>` when `max-width: 760px` since `margin-left` and `margin-right` are added.

Screen Capture on 2022-11-21 at 03-14-38

renintw (approved):

StevenDufresne:
> Tested and looks good 👍
> 
> One thing: Can we somehow make the text `email` be on the third line on mobile devices? image Perhaps adding `margin: 0` for `<p>` when `@media (max-width: 760px)` as `margin-left` and `margin-right` are added to it and its parent.

I would prefer not using any custom CSS for simple pages like this. I tried a bunch of different techniques through the editor but since we have CSS rules that add margin-left & margin-right for mobile layouts, it didn't work well with the `max-width` set to control the `<p>` width at a certain breakpoint.

I think I'll merge this as is.

PR #50, opened 2022-11-21
Query Modifications.

StevenDufresne:
This PR solves 2 problems:
1. Pages are returning in search results
2. Sticky posts show up twice in paginated results.

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/functions.php
+++ b/source/wp-content/themes/wporg-showcase-2022/functions.php
@@ -20,7 +20,6 @@ add_filter( 'excerpt_length', __NAMESPACE__ . '\modify_excerpt_length', 999 );
 add_filter( 'excerpt_more', __NAMESPACE__ . '\modify_excerpt_more' );
 add_filter( 'query_loop_block_query_vars', __NAMESPACE__ . '\modify_query_loop_block_query_vars', 10, 2 );
 
-
 /**
  * Enqueue scripts and styles.
  */
--- a/source/wp-content/themes/wporg-showcase-2022/functions.php
+++ b/source/wp-content/themes/wporg-showcase-2022/functions.php
@@ -180,7 +180,7 @@ function modify_query_loop_block_query_vars( $query, $block ) {
 /**
  * We should only return pages from searches within this theme.
  *
- * @return void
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-showcase-2022/functions.php
```diff
@@ -175,3 +176,13 @@ function modify_query_loop_block_query_vars( $query, $block ) {
...
+
+/**
+ * We should only return pages from searches within this theme.
+ *
+ * @return void
+ */
+function remove_pages_from_search() {
+    global $wp_post_types;
+    $wp_post_types['page']->exclude_from_search = true;
+}
```
dd32, line 188: This should probably be done via `register_{$post_type}_post_type_args` instead of altering the global directly.

dd32 (commented):
I can't see any reason why pages should be included in the search for the showcase.

Normally I'd suggest using a `pre_get_posts` filter, checking for `is_search()` and setting `post_type` to the set of types that should be searched... but setting the post_type `exclude_from_search` param seems a reasonable approach as well.

source/wp-content/themes/wporg-showcase-2022/functions.php
StevenDufresne: > I can't see any reason why pages should be included in the search for the showcase.
> 
> Normally I'd suggest using a `pre_get_posts` filter, checking for `is_search()` and setting `post_type` to the set of types that should be searched... but setting the post_type `exclude_from_search` param seems a reasonable approach as well.

Yeah, that sounds better. Will update.

StevenDufresne: Done [c98b97b](https://github.com/WordPress/wporg-showcase-2022/pull/50/commits/c98b97b9aaec60bb5a5e9df48818f1bb8451f37c)

StevenDufresne: Do i need to check for `$query->is_main_query()`?

dd32: Yeah, best to check for `is_main_query`. LGTM Steve.

PR #57, opened 2022-11-23
Forces users to log in before viewing site submission page.

### LINKED ISSUE #53: You should be logged in to submit a site.
Accessing `/submit-a-wordpress-site` should alert the user to log in before they can view that page.

# Designs
image

### LINKED ISSUE #54: Add redirect for sites that are have been removed from directory.
We will not launch with all the current sites. We need to redirect all existing sites to the homepage.

StevenDufresne:
Fixes #53.

This PR adds a redirect to make sure that only logged-in users can visit the site submission page. 

## Approach
This adds a template `page-submit-auth` which makes the template editable for moderators. If a user tries to visit `submit-a-wordpress-site` and is not logged in, we load said template.

## Alternate approach
We could create a front-end page and use this template and redirect instead of loading the template but it's probably more data management than it's worth. Also, it would be nice if the JetPack form builder has an auth feature, but it doesn't.

## Screenshots

| Mobile | Tablet | Desktop |
| --- | --- | --- |
| localhost_8888_submit-a-wordpress-site_(iPhone 12 Pro) | localhost_8888_submit-a-wordpress-site_(iPad Air) | localhost_8888_submit-a-wordpress-site_ (2) |

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/patterns/page-submit-auth.php
+++ b/source/wp-content/themes/wporg-showcase-2022/patterns/page-submit-auth.php
@@ -7,18 +7,19 @@
 
 ?>
 
-<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"}}}} -->
-<div class="wp-block-group" style="padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60)"><!-- wp:heading -->
-<h1><?php esc_html_e( 'Before you submit...', 'wporg' ); ?></h1>
+<!-- wp:group {"style":{"spacing":{"blockGap":"0px","padding":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70"}}}}} -->
+<div class="wp-block-group" style="padding-top:var(--wp--preset--spacing--70);padding-bottom:var(--wp--preset--spacing--70)"><!-- wp:heading {"textAlign":"center","style":{"spacing":{"margin":{"top":"var:preset|spacing|40","bottom":"var:preset|spacing|20"}}},"fontSize":"heading-3"} -->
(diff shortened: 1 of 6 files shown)


javierarce:
I'll get you a design for this. 

By the way, I think this PR is referencing the wrong issue (#54 instead of #53)

thetinyl:
Thanks for the tag @javierarce! Copy reviewed and made some suggestions in the [Figma file](https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA?node-id=1736:23024#314165617).

source/wp-content/themes/wporg-showcase-2022/patterns/page-submit-auth.php
```diff
@@ -0,0 +1,25 @@
...
+<!-- /wp:heading -->
+
+<!-- wp:group {"layout":{"type":"constrained","contentSize":"480px"}} -->
+<div class="wp-block-group"><!-- wp:paragraph {"align":"center"} -->
+<p class="has-text-align-center">
+<?php
+printf(
+	/* translators: %s is the login url. */
+	wp_kses_post( __( 'Thanks for your interest in submitting a site for the showcase! In order to do so, please <a href="%s">login to your WordPress.org account</a>.', 'wporg' ) ),
+	esc_url( wp_login_url( home_url() . $_SERVER['REQUEST_URI'] ) )
```
StevenDufresne, line 22: @tellyworth @dd32 Since I'm redirecting to the template, there's no wp-query to pull URLs from. Is this a good idea?

StevenDufresne: I actually wasn't able to get this past the linter so i just gave up :)

outdoor2kode (approved):
Tested and looks good 👍

javierarce:
I've updated [the design](https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=1736%3A23024&t=j0td3retIOgRvAUA-4) with @thetinyl recommendations:

- Title: Log in to submit a site
- Body: Thanks for your interest in submitting a site to the Showcase! [Log in to your WordPress.org account](#) to continue.

image

PR #58, opened 2022-11-23
Set no index if we are on the thanks page.

### LINKED ISSUE #52: Page `/thanks` should be noindex.

StevenDufresne:
Fixes #52 

We have a confirmation page that a user is redirected to after they submit a site. This PR sets that page to be no index. Additionally, there shouldn't be any other references to that page in the theme.

# How to test
You'll need a sandbox
 
- Apply patch to sandbox
- Visit `wordpress.org/showcase-test/submit-a-wordpress-site/thanks`
- Expect to see `x-robots: noindex,follow` returned in the response.

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/functions.php
+++ b/source/wp-content/themes/wporg-showcase-2022/functions.php
@@ -231,7 +231,7 @@ function redirect_urls() {
 	}
 }
 
-/**
+/**	
  * Robots "noindex" rules for specific parts of the Learn site.
  *
  * @param bool $noindex
--- a/source/wp-content/themes/wporg-showcase-2022/functions.php
+++ b/source/wp-content/themes/wporg-showcase-2022/functions.php
@@ -231,7 +231,7 @@ function redirect_urls() {
 	}
 }
 
-/**	
+/**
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-showcase-2022/functions.php
```diff
@@ -200,3 +201,21 @@ function modify_search_query( $query ) {
...
+ * Robots "noindex" rules for specific parts of the showcase site.
+ *
+ * @param bool $noindex
+ *
+ * @return bool
+ */
+function set_noindex( $noindex ) {
+	global $pagename;
+
+	if ( 'thanks' == $pagename ) {
```
adamwoodnz, line 215: Should this use `===`?

adamwoodnz (approved):
One question, but works for me

source/wp-content/themes/wporg-showcase-2022/functions.php
StevenDufresne: That's a good question. I imagine a strict comparison is always preferred. I'm surprised this isn't a standard rule. I'll update.

PR #60, opened 2022-11-24
Update the page-submit callout on the front page.

StevenDufresne:
This PR aligns the submit callout to the left side.

Notes: changes agreed upon here need to be ported to the same component on the homepage.
[Parent pattern link](https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php).

## ScreenShots
Screen Shot 2022-11-24 at 2 15 04 PM

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/patterns/front-page-submit.php
+++ b/source/wp-content/themes/wporg-showcase-2022/patterns/front-page-submit.php
@@ -8,7 +8,7 @@
 ?>
 
 <!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"spacing":{"padding":{"top":"80px","bottom":"80px"}}},"backgroundColor":"charcoal-2","textColor":"white","layout":{"inherit":true,"type":"constrained"}} -->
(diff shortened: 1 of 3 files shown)


adamwoodnz (commented):
Works for me above 760px wide. Below that the left alignment breaks:

Screen Shot 2022-11-28 at 11 21 04 AM

Not sure if that's intentional.

StevenDufresne:
> Not sure if that's intentional.

Thanks for the review. It's not, I was having difficulting getting the `max-width` to apply properly, and it looks like I flipped the problem on its head late friday :). Fixed the desktop, but broke mobile 😵 . 

The problem is that we use some inferred layout controls to add padding, introduced in https://github.com/WordPress/wporg-showcase-2022/pull/34. However, it falls apart when you try to use Gutenberg's max-width functionality because it also adds padding & margin to the specific container via php. In order to get this working, I had to move to empty columns. 🤔

Mind giving it another look?

adamwoodnz:
Works for me now. I tried to get something working without the empty columns by setting the H2 and paragraph below to `alignleft` to stop the margins being applied, and this worked in the browser but the editor blew up as the blocks were invalid 🤷 

If we're doing everything in the editor I don't see a better way.

adamwoodnz (approved):

StevenDufresne:
> Works for me now. I tried to get something working without the empty columns by setting the H2 and paragraph below to `alignleft` to stop the margins being applied, and this worked in the browser but the editor blew up as the blocks were invalid 🤷

Oh, that's an interesting idea. Let me try that, if I can get it to work, I'll replace, otherwise merge.

StevenDufresne:
> > Works for me now. I tried to get something working without the empty columns by setting the H2 and paragraph below to `alignleft` to stop the margins being applied, and this worked in the browser but the editor blew up as the blocks were invalid 🤷
> 
> Oh, that's an interesting idea. Let me try that, if I can get it to work, I'll replace, otherwise merge.

Right, those components don't support those layout options. 😞

Issue #59, opened 2022-11-24
Design Audit Tasks


StevenDufresne:
#### [Design team feedback threads](https://nesosoup.wordpress.com/2022/11/22/showcase-redesign-quick-update-2/#comment-2293)

**Tasks**
- [x] Round images to `2px` if not full width
- [x] Round Single site meta details 
- [x] Left align Submit section, do same for .org homepage (https://github.com/WordPress/wporg-showcase-2022/pull/60)
- [x] Expand content width to match designs
- [x] Reduce Size of subnav text
- [x] Remove paint brush stroke to sub-nav -- Owner: @REN
- [ ] Fix fluid alignment of content & surrounding components (https://github.com/WordPress/wporg-showcase-2022/issues/62)
- [x] Reduce pagination link size.

StevenDufresne:
Closing as the last item remaining is logged as a separate issue.

PR #63, opened 2022-11-28
Design Audit: Reduce subnav text size

outdoor2kode:
See https://github.com/WordPress/wporg-showcase-2022/issues/59

## Screenshots

image

image

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-breadcrumbs/style.scss
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-breadcrumbs/style.scss
@@ -1,7 +1,6 @@
 .wp-block-wporg-site-breadcrumbs > div {
 	display: flex;
 	align-items: center;
-	min-height: var(--local-header-height);
 
 	a {
 		color: inherit;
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-showcase-2022/src/site-breadcrumbs/style.scss
```diff
@@ -1,7 +1,7 @@
 .wp-block-wporg-site-breadcrumbs > div {
 	display: flex;
 	align-items: center;
-	min-height: var(--local-header-height);
+	font-size: var(--wp--preset--font-size--small);
```
ryelle, line 4: Rather than set this here, this setting is available on the block itself so it should be changed in `patterns/subnav-bar.php`:

```diff
- <!-- wp:wporg/site-breadcrumbs {"textColor":"charcoal-1"} /-->
+ <!-- wp:wporg/site-breadcrumbs {"textColor":"charcoal-1","fontSize":"small"} /-->
```

renintw: Ah, thanks! 👍

ryelle (approved):

PR #64, opened 2022-11-28
Design Audit: Remove paintbrush stroke

renintw:
See #59 

~At first, I was trying to add `.is-sticky` to [parent-2021](https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss#L394), but I don't think that makes sense as the attributes assigned there should only be applied to `.is-style-brush-stroke.is-sticky` rather than `.is-sticky`. I'm thinking there might be other forms of `sticky` child theme would like to use, so I add it here first.~

~The mobile dropdown is broken after removing `is-style-brush-stroke`. Fixed [here](https://github.com/WordPress/wporg-parent-2021/pull/60).~

**Update**: Check out https://github.com/WordPress/wporg-parent-2021/pull/60#issuecomment-1331133075

## Screencast

https://user-images.githubusercontent.com/18050944/204383655-4a6166f6-90aa-4241-9a3f-2331581cad7a.mp4

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/patterns/subnav-bar.php
+++ b/source/wp-content/themes/wporg-showcase-2022/patterns/subnav-bar.php
@@ -7,8 +7,8 @@
 
 ?>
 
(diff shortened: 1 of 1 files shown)


ryelle:
I've asked in slack about when the brush stroke style should be used, so it's possible we'll end up changing how `is-brush-stroke-style` and `is-sticky` interact in the parent theme. I'd hold off on working on this until there's an answer.

PR #65, opened 2022-11-29
Blocks: Remove Site Breadcrumbs block code

ryelle:
See https://github.com/WordPress/wporg-mu-plugins/issues/304, https://github.com/WordPress/wporg-mu-plugins/pull/305 — This moves the Site Breadcrumbs block to wporg-mu-plugins, so that we can use it on other sites too.

This PR removes the local `wporg/site-breadcrumbs` block, and since the wporg-mu-plugins one uses the same name, nothing needs to change. The colors on the group block were a little off, so I've fixed that too.

The new block uses `get_bloginfo( 'name', 'display' )` which for Showcase seems to be "The WordPress Showcase," so I'm using the new filter to update that to just "Showcase."

Screenshot 2022-11-28 at 6 46 23 PM

Screenshot 2022-11-28 at 6 48 29 PM

To test

- Apply the wporg-mu-plugins PR https://github.com/WordPress/wporg-mu-plugins/pull/305
- Load up this branch
- There should be no visual changes


StevenDufresne (approved):
I ran this locally, working as expected.

renintw (commented):
Tested and looks good 👍

PR #66, opened 2022-11-30
Block Style: Move the secondary search style to the parent theme

ryelle:
See https://github.com/WordPress/wporg-parent-2021/pull/61 — this style is moving to the parent theme, so it can be removed here.

Test in combination with the parent theme PR (or once the parent theme PR is merged), and there should be no change to the search field.

Screenshot 2022-11-30 at 1 41 27 PM


StevenDufresne (approved):
Still working as expected 👍

renintw (approved):
Tested and looks good 💯

Issue #31, opened 2022-11-03
Ensuring consistent, high-quality screen shots.


StevenDufresne:
The goal of this ticket to align on the following:
1. What are the aspect ratios?
2. How can we maintain consistency between mShot images and uploaded screenshots

## What are the aspect ratios?
There are 3 different places where images are displayed:
- Single Template:
  - [Main header](https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=850%3A17307)
  - [Related Sites](https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=850%3A17307)
- [Archive views](https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=927%3A13282)

Ideally, these all share the same aspect ratio so when moderators upload screenshots, they only need to upload 1 high-resolution image. Since our designs aren't incredibly image-heavy, I don't think performance will be really impacted.

**What is the ideal aspect ratio?**

## How can we maintain consistency between mShot images and uploaded screenshots.
While I don't think we should go so far as validating the images or resizing them, I think we can add the expected dimensions in the custom field display. 

Something like:
<img width="976" alt="Screen Shot 2022-11-03 at 12 25 40 PM" src="https://user-images.githubusercontent.com/1657336/199642130-0aa988bf-4b92-41aa-8477-79deb99ba39c.png">

If we do so, we can use CSS aspect ratio and assume that uploaded images will be the correct dimensions.

**What would be the ideal resolution?**

jasmussen:
Good issue, excellent screenshots is kind of make or break for the showcase site. Kind of like a cookbook, the "nice pictures" is a primary reason to visit. Large high quality screenshots showcasing the site in its best light:

* I would think sizes be _at least_ 2880x1620 but ideally even larger, such as 3840x2160. We can compensate by highly jpeg compressing, for example at quality 25 or 30, which usually can work well with high resolution screenshots.
* Screenshots should never show cookie notices, or cloudflare firewalls, "subscribe to our newsletter" modals, or any errors.
* Ideally we can avoid showing ads on the screenshot. Sites with ads on the homepage above the fold might even be something to consider omitting from the directory entirely.

StevenDufresne:
Okay, so let's agree that image screenshots will have a ratio: `16:9`. For now, let's also assume all images have the same ratio 👍 .

[...]

PR #67, opened 2022-12-01
Stop loading large thumbnails.

StevenDufresne:
This PR introduces the ability to set the site-screenshot block to download a high resolution image, improving page performance. Currently, all images are loading with `scale=2` applied which doubles the size of the image. That resolution is only needed for the homepage hero image and the single view.

# Considerations
I played with using src-set to only use high-resolution images for set screenshots on `>= tablet` in [this branch](https://github.com/WordPress/wporg-showcase-2022/compare/fix/smaller-thumnail-images-srcset?expand=1), however, mShots considers them different photos and for some images there could be visible differences in the image.

**Example**

## How to test
1. Make sure you have cleared your front-page customizations
2. Load the front-page with Dev Tools network tab open (filter, search `scale=2`)
3. Expect to only see 1 image loaded with that query string
4. Navigate to archive, expect so see none
5. Click on site in archive, expect to see 1 image loaded.

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
@@ -51,7 +51,7 @@ function render( $attributes, $content, $block ) {
 		$screenshot = add_query_arg( 'scale', 2, $screenshot );
 	}
 
-	$img_content = "<img src='{$screenshot}' alt='" . the_title_attribute( array( 'echo' => false ) ) . "' loading='lazy' />";
+	$img_content = "<img src='" . esc_url( $screenshot ) . "' alt='" . the_title_attribute( array( 'echo' => false ) ) . "' loading='lazy' />";
 
 	if ( isset( $attributes['isLink'] ) && true == $attributes['isLink'] ) {
 		$img_content = '<a href="' . get_permalink( $post ) . '">' . $img_content . '</a>';
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
@@ -51,8 +51,6 @@ function render( $attributes, $content, $block ) {
 		$screenshot = add_query_arg( 'scale', 2, $screenshot );
 	}
 
-	$screenshot = add_query_arg( 'scale', 0.5, $screenshot );
-
 	$img_content = "<img src='{$screenshot}' alt='" . the_title_attribute( array( 'echo' => false ) ) . "' loading='lazy' />";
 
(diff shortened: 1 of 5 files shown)


source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
```diff
@@ -46,7 +46,10 @@ function render( $attributes, $content, $block ) {
 	$post = get_post( $post_ID );
 
 	$screenshot = site_screenshot_src( $post );
-	$screenshot = add_query_arg( 'scale', 2, $screenshot );
+
+	if ( isset( $attributes['useHiRes'] ) && true === $attributes['useHiRes'] ) {
+		$screenshot = add_query_arg( 'scale', 2, $screenshot );
+	}
 
 	$img_content = "<img src='{$screenshot}' alt='" . the_title_attribute( array( 'echo' => false ) ) . "' loading='lazy' />";
```
adamwoodnz, line 54: [Sounds like this should be escaped](https://a8c.slack.com/archives/G010HCMN0HF/p1669942609107009)

```suggestion
		$screenshot = add_query_arg( 'scale', 2, $screenshot );
	}

	$img_content = "<img src='" . esc_url( $screenshot ) . "' alt='" . the_title_attribute( array( 'echo' => false ) ) . "' loading='lazy' />";
```

adamwoodnz:
> for some images there could be visible differences in the image

Would this actually impact a user? Normal users don't often resize their browser (unless rotating a smaller device), so I wonder whether it's an issue. It troubles me that on a mobile device we'll still be loading an image which is 2880px wide...

Screen Shot 2022-12-02 at 2 20 58 PM

StevenDufresne:
> > for some images there could be visible differences in the image
> 
> Would this actually impact a user? Normal users don't often resize their browser (unless rotating a smaller device), so I wonder whether it's an issue. It troubles me that on a mobile device we'll still be loading an image which is 2880px wide...

That is a good point. Device rotation is the main use case and would affect nearly all devices (very likely? probably not). Something to consider is that mShots returns a loading image if there isn't an image in the cache. So you may go from having an image to seeing a loader on rotation. 

In addition, it's more than likely that we will be using [already saved screenshots,](https://github.com/WordPress/wporg-showcase-2022/issues/31) compressed and converted to `webp` automatically by jetpack. In those cases, we won't have different assets to load. There will only ever be 1 screenshot saved. As it stands today, we won't be creating thumbnails from the manually uploaded screenshots. 

With all that in mind, I think I'm still comfortable with this PR. What about you?

adamwoodnz (approved):
The image sizing and format is definitely affecting perf scores at the moment, particularly on mobile, but if we move to webp that will have a significant effect.

Screen Shot 2022-12-06 at 10 01 32 AM

👍 So I think this is a step in the right direction.

Issue #68, opened 2022-12-05
Related Sites title should be hidden if there are none.


StevenDufresne:
The JetPack `Related sites` block doesn't include a title so if there are no related sites, we get a title but then no content under it. ([Relevant Theme Code](https://github.com/WordPress/wporg-showcase-2022/blob/ea3af4bc9581d68145eef301aff9e607201f82e8/source/wp-content/themes/wporg-showcase-2022/patterns/page-single.php#L25))

Should we:
1. Try to add a PR for JetPack
2. Use some code in our theme to hide the title.
    - [We could use a filter and hide it](https://jetpack.com/support/related-posts/customize-related-posts/). 
3. Could use `has` CSS property
    - `h3:has(+ .wp-block-jetpack-related-posts){ display: block; }`  
    - No FireFox support

See: https://github.com/Automattic/jetpack/issues/27108

## Example
<img width="887" alt="Screen Shot 2022-12-05 at 9 28 10 AM" src="https://user-images.githubusercontent.com/1657336/205525332-61b804a2-7d96-4483-bc95-2c7b680fde30.png">

StevenDufresne:
I'm moving this to a Future milestone seeing that we have a temporary fix that works.

PR #69, opened 2022-12-05
Hide the title if there are no related sites.

StevenDufresne:
This temporarily hides the `related posts` title while a better, long-term fix is worked on for https://github.com/WordPress/wporg-showcase-2022/issues/68.

The drawback here is that if more content is added in the container, this will break. For that reason, this cannot be considered a long-term fix. The best case is that the block itself handles the displaying/hiding of the title.

## Alternative CSS approaches
We could also use `h3:has( + { jetpack_container_id } )` but it's not supported in firefox. `only-child` has `97.49%` support according to caniuse.com, making it a better approach. 

## How to test
Need a sandbox.

1. Navigate to a site with no related posts (https://wordpress.org/showcase-test/2009/01/25/oregon-coast-visitors-association/), expect to **not** see the related posts title.
2. Navigate to other sites, and expect to see the related posts title. (https://wordpress.org/showcase-test/2009/03/23/we-heart-this/)

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/patterns/page-single.php
+++ b/source/wp-content/themes/wporg-showcase-2022/patterns/page-single.php
@@ -23,9 +23,11 @@
 <!-- /wp:group -->
 
 <!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|40"}}}} -->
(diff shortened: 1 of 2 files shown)


StevenDufresne:
🤣 I love assigning/unassigning.

adamwoodnz:
Works for me and seems a reasonable approach, given that I wouldn't expect anything unrelated to be added to this container :shipit:

adamwoodnz (approved):

Issue #71, opened 2022-12-05
Add title to submit page.


StevenDufresne:
We need an `<h1>` on the submit a site page.

## Screenshot
Link: https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=1771%3A24626&t=QUmtn8DDdE9dAuOU-0

![image](https://user-images.githubusercontent.com/1657336/205529379-57f29c9c-011a-4934-875f-c32ef15f0919.png)

PR #73, opened 2022-12-05
Fix heading levels.

### LINKED ISSUE #72: Audit Page heading structure.

StevenDufresne:
Closes #72.

This PR fixes heading levels for the following pages:
- Front-page
- Single
- Thanks

### Front-page
Page: https://wordpress.org/showcase-test
| Before | After | 
| --- | --- |
| Screen Shot 2022-12-05 at 10 37 42 AM | Screen Shot 2022-12-05 at 10 54 57 AM |

### Single 
Page: https://wordpress.org/showcase-test/2019/10/07/vogue/
| Before | After | 
| --- | --- |
|  Screen Shot 2022-12-05 at 11 04 19 AM | Screen Shot 2022-12-05 at 11 04 32 AM |

###  Thanks
Page: https://wordpress.org/showcase-test/submit-a-wordpress-site/thanks
| Before | After | 
| --- | --- |
| Screen Shot 2022-12-05 at 11 21 52 AM | Screen Shot 2022-12-05 at 11 21 37 AM |

### Archives
We don't have a good candidate for an `<h1>` on archives though. We could make the context statement an `<h1>` I guess. 

**Any opinions?**

#### Current Page
Screen Shot 2022-12-05 at 11 25 15 AM

## Submit Site 
This should be fixed when https://github.com/WordPress/wporg-showcase-2022/issues/71 is closed.

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/patterns/front-page.php
+++ b/source/wp-content/themes/wporg-showcase-2022/patterns/front-page.php
@@ -13,7 +13,7 @@
 <div class="wp-block-group has-charcoal-2-background-color has-background"><!-- wp:columns {"align":"full","style":{"spacing":{"blockGap":{"top":"var:preset|spacing|40","left":"var:preset|spacing|40"},"padding":{"top":"0"}}}} -->
 <div class="wp-block-columns alignfull" style="padding-top:0"><!-- wp:column {"width":"60%","layout":{"type":"default"}} -->
 <div class="wp-block-column" style="flex-basis:60%"><!-- wp:heading {"level":1,"className":"screen-reader-text"} -->
-<h1 class="screen-reader-text"><?php esc_attr_e( 'Showcase:', 'wporg' ); ?></h1>
+<h1 class="screen-reader-text"><?php esc_attr_e( 'Showcase', 'wporg' ); ?></h1>
 <!-- /wp:heading -->
 
 <!-- wp:group {"style":{"spacing":{"padding":{"right":"var:preset|spacing|30"}}},"layout":{"type":"constrained"}} -->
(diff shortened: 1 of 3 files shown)


adamwoodnz:
> We could make the context statement an <h1> I guess

An a11y expert would know more, but I feel like this would be valid

source/wp-content/themes/wporg-showcase-2022/patterns/front-page.php
```diff
@@ -28,11 +28,11 @@
...
-<p class="has-text-align-left is-style-serif has-white-color has-text-color" style="line-height:0"><?php esc_attr_e( 'Featuring:', 'wporg' ); ?></p>
-<!-- /wp:paragraph -->
+<div class="wp-block-group"><!-- wp:heading {"textAlign":"left","style":{"typography":{"lineHeight":"0"}},"textColor":"white","className":"is-style-serif","fontSize":"extra-large"} -->
+<h2 class="has-text-align-left is-style-serif has-white-color has-text-color has-extra-large-font-size" style="line-height:0"><?php esc_attr_e( 'Featuring:', 'wporg' ); ?></h2>
+<!-- /wp:heading -->
 
-<!-- wp:post-title {"textAlign":"left","isLink":true,"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"typography":{"fontStyle":"italic","fontWeight":"400"}},"textColor":"white","fontSize":"extra-large","fontFamily":"eb-garamond"} /--></div>
+<!-- wp:post-title {"level":3,"textAlign":"left","isLink":true,"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"typography":{"fontStyle":"italic","fontWeight":"400"}},"textColor":"white","fontSize":"extra-large","fontFamily":"eb-garamond"} /--></div>
```
adamwoodnz, line 35: Mixed feelings about this one. Feels like a sentence to me so maybe should all be one H2

StevenDufresne: Yeah, it's tricky because we use the `post-title`, which I think makes sense from a reusability standpoint. Maybe there's a way to target that post-title and filter it. 

On the other hand, since we do use the query block, it ends up being a list of one item with the same structure as the features list section. In that regard, it is consistent with the grid below.

StevenDufresne:
> > We could make the context statement an `<h1>` I guess
> 
> An a11y expert would know more, but I feel like this would be valid

I thinking about merging this once we resolve your feedback [here](https://github.com/WordPress/wporg-showcase-2022/pull/73#discussion_r1040356119). Then deploying, splitting the archive page into a new ticket, and then pinging the a11y team. 

I feel we'll get better feedback with more code in place. Thoughts?

source/wp-content/themes/wporg-showcase-2022/patterns/front-page.php
adamwoodnz: Yeah I see now. Still getting my head around building with the site editor. Definitely introduces limitations as I'm finding with the 'No search results' page.

My understanding is that a lot of screen reader users will navigate through the headings to get a sense of the page content, and I'm still a little unsure about how this will work, but I guess it will just read them consecutively, even if there is a pause. Maybe just go with it and seek a11y feedback once it's launched as you say.

PR #79, opened 2022-12-05
Add better social media meta data.

StevenDufresne:
Using the [pattern directory code as a guide](https://github.com/WordPress/pattern-directory/blob/trunk/public_html/wp-content/themes/pattern-directory/functions.php#L240), this PR adds better meta data for SEO & social embeds.

## Examples
| Page | Facebook | Twitter |
| --- | --- | --- |
| Front Page, Tags, Search | Screen Shot 2022-12-05 at 3 48 25 PM | Screen Shot 2022-12-05 at 3 48 20 PM |
| Single | Screen Shot 2022-12-06 at 10 10 22 AM | Screen Shot 2022-12-06 at 10 10 33 AM  |

I considered using the site thumbnail for the `single` image but it felt like we lost context. Here's is an example of what it looked like:
Screen Shot 2022-12-05 at 3 56 07 PM

I can go either way on that. **Thoughts?**

**RESOLUTION: Added site thumbnail.**

@jasmussen @javierarce can I get a default social image to replace the one in this pull request?


javierarce:
> I considered using the site thumbnail for the single image but it felt like we lost context. 

I think using the actual thumbnail is a good idea and is a common practice. Showing a generic image would be strange.

StevenDufresne:
> > I considered using the site thumbnail for the single image but it felt like we lost context. Here's is an example of what it looked like:
> 
> That's perfect. I think using the actual thumbnail is a good idea and is a common practice. Showing a generic image would be strange.
> 
> > can I get a default social image to replace the one in this pull request?
> 
> Yes, I'll design something for this

Thanks. Added the thumbnail (See issue description above). Awaiting the default image thumbnail. 👍

source/wp-content/themes/wporg-showcase-2022/functions.php
```diff
@@ -316,3 +317,46 @@ function document_title( $parts ) {
...
+ */
+function add_social_meta_tags() {
+	$default_image = get_stylesheet_directory_uri() . '/images/social-image.png';
+	$site_title    = function_exists( '\WordPressdotorg\site_brand' ) ? \WordPressdotorg\site_brand() : 'WordPress.org';
+	$og_fields = [
+		'og:title'       => wp_get_document_title(),
+		'og:description' => __( 'Discover inspiration in some of the most beautiful, best designed WordPress websites.', 'wporg' ),
+		'og:site_name'   => $site_title,
+		'og:type'        => 'website',
+		'og:url'         => home_url(),
```
dd32, line 332: This should use a trailing-slashed URL:
```suggestion
		'og:url'         => home_url('/'),
```

source/wp-content/themes/wporg-showcase-2022/functions.php
```diff
@@ -316,3 +317,46 @@ function document_title( $parts ) {
...
+
+	foreach ( $og_fields as $property => $content ) {
+		printf(
+			'<meta property="%1$s" content="%2$s" />' . "\n",
+			esc_attr( $property ),
+			esc_attr( $content )
+		);
+	}
+
+	if ( isset( $og_fields['og:description'] ) ) {
```
dd32, line 356: I'm not sure if this should be outputting both `<meta property="og:description"` and `<meta name="description"` I thought it should only be the latter.. But both works for me if we do that elsewhere.

dd32 (approved):

source/wp-content/themes/wporg-showcase-2022/functions.php
StevenDufresne: We do it in the loop above. 👍.

StevenDufresne:
@javierarce 

I've opened a new issue for replacing the placeholder image: 
https://github.com/WordPress/wporg-showcase-2022/issues/83.

Merging this since the code is approved.

source/wp-content/themes/wporg-showcase-2022/functions.php
dd32: Yeah, I was saying I'm not sure it should output both.

We output both on plugins, so this is fine.
```
$ curl https://wordpress.org/plugins/hello-dolly/ -s | grep '<meta.\+description'
<meta property="og:description" content="This is not just a plugin, it symbolizes the hope and enthusiasm of an entire generation summed up in two words sung most famously by Louis Armstrong." />
<meta name="description" content="This is not just a plugin, it symbolizes the hope and enthusiasm of an entire generation summed up in two words sung most famously by Louis Armstrong." />
```

Issue #75, opened 2022-12-05
Allow the ability to pass lazy loading to site screenshot.


StevenDufresne:
Based on a lighthouse performance review, if we know that the image will appear in the viewport on load, it shouldn't have lazy loading applied. That will lead to a delay in painting. 

[Relevant source code](https://github.com/WordPress/wporg-showcase-2022/blob/0a6c76a10b0e6f175f4abf171a46d03e5af0e624/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php#L51).

We have 2 large images that are loaded immediately that we should update:
- https://github.com/WordPress/wporg-showcase-2022/blob/0a6c76a10b0e6f175f4abf171a46d03e5af0e624/source/wp-content/themes/wporg-showcase-2022/patterns/front-page.php#L40
- https://github.com/WordPress/wporg-showcase-2022/blob/0a6c76a10b0e6f175f4abf171a46d03e5af0e624/source/wp-content/themes/wporg-showcase-2022/templates/single.html#L4

<img width="543" alt="Screen Shot 2022-12-05 at 1 04 08 PM" src="https://user-images.githubusercontent.com/1657336/205547019-fd453477-2df7-4381-94cf-07d723b64a5f.png">

PR #82, opened 2022-12-06
Update site-screenshot to allow setting lazy loading.

StevenDufresne:
Closes: #75 

Currently, all images that use `site-screenshot` are lazy loaded. This delays the first paint of the page needlessly. We should not have it lazy loading on images that we are sure will show up on the viewport on load. We have 2 such images -- front page, single. 

This PR makes lazy loading a property to pass into the block, assuming default as the initial state, matching the default `` behavior. 

**Performance**
Reduces `Largest Contentful Paint` on average by `~300ms`

**Block Editor View**
Screen Shot 2022-12-06 at 11 21 03 AM


adamwoodnz:
Works for me, and default makes sense :shipit:

adamwoodnz (approved):

PR #85, opened 2022-12-06
Add layout for no search results

### LINKED ISSUE #70: Implement No Search Results Design
When there are no search results we need to display something better.

## Screenshot
Link: https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=1809%3A23682&t=QUmtn8DDdE9dAuOU-0

image

adamwoodnz:
Closes #70 

Adds a custom layout for the search results page when no sites are found.

Not rendering the standard results bar at the top of the page is tricky because there is no way to conditionally render it using block markup, I have resorted to using CSS to not display it. See additional comments inline. Please let me know if there's a better way.

| Desktop | Mobile |
|-|-|
| Screen Shot 2022-12-07 at 1 02 33 PM | Screen Shot 2022-12-07 at 1 03 03 PM |

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,9 +20,7 @@
 		"backstop:test": "npm exec backstop test",
 		"lighthouse": "lighthouse-ci http://localhost:8888/ --accessibility=100 --best-practices=100 --seo=100",
 		"lighthouse:desktop": "lighthouse http://localhost:8888/ --view --preset=desktop --output-path=lighthouse.html",
-		"lighthouse:mobile": "lighthouse http://localhost:8888/ --view --screenEmulation.mobile --output-path=lighthouse.html",
-		"lint:theme": "composer run lint source/wp-content/themes/wporg-showcase-2022 && yarn workspace wporg-showcase-2022-theme lint:js && yarn workspace wporg-showcase-2022-theme lint:css",
-		"watch:theme": "yarn workspace wporg-showcase-2022-theme start"
+		"lighthouse:mobile": "lighthouse http://localhost:8888/ --view --screenEmulation.mobile --output-path=lighthouse.html"
 	},
 	"workspaces": [
 		"source/wp-content/themes/wporg-showcase-2022"
--- a/package.json
+++ b/package.json
@@ -20,7 +20,9 @@
 		"backstop:test": "npm exec backstop test",
 		"lighthouse": "lighthouse-ci http://localhost:8888/ --accessibility=100 --best-practices=100 --seo=100",
 		"lighthouse:desktop": "lighthouse http://localhost:8888/ --view --preset=desktop --output-path=lighthouse.html",
-		"lighthouse:mobile": "lighthouse http://localhost:8888/ --view --screenEmulation.mobile --output-path=lighthouse.html"
+		"lighthouse:mobile": "lighthouse http://localhost:8888/ --view --screenEmulation.mobile --output-path=lighthouse.html",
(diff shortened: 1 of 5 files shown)


source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
```diff
@@ -142,3 +142,7 @@
 		}
 	}
 }
+
+.search-no-results .wporg-archive-results-bar {
+	display: none;
+}
```
adamwoodnz, line 148: I couldn't find a way to conditionally render this if there were sites found. There is a `<!-- wp:query-no-results -->` block, but not a `<!-- wp:query-results -->` block, and the layout is too different between the two. Hiding the top one based on the body class is the best I can come up with. Any better solution?

Results:
Screen Shot 2022-12-07 at 12 56 24 PM

No results:
Screen Shot 2022-12-07 at 12 56 03 PM

StevenDufresne:
Could we server up a different template where there are no results?

StevenDufresne:
After playing with this experience, I'm not certain I like that the search bar moves locations when there are no results. It seems like extra cognitive load for little return. Thoughts?

@javierarce, do we need to serve up a different template for no results? Can we do something similar to the pattern directory, albeit nicer looking?

Screen Shot 2022-12-07 at 9 40 44 AM

adamwoodnz:
> Could we serve up a different template when there are no results?

Nothing seems to work based on the typical hierarchy. `search-no-results.html` didn't load

adamwoodnz:
Updated to match screenshot above

source/wp-content/themes/wporg-showcase-2022/patterns/archive-content.php
```diff
@@ -10,31 +10,39 @@
...
+
 	<!-- wp:query-pagination-next /-->
 	<!-- /wp:query-pagination -->
-	
+
 	<!-- wp:query-no-results -->
-	<!-- wp:paragraph {"placeholder":"Add text or blocks that will display when a query returns no results."} -->
-	<p><?php esc_attr_e( 'No results.', 'wporg' ); ?></p>
+	<!-- wp:paragraph {"align":"center"} -->
+	<p class="has-text-align-center">View <a href="/archives" data-type="URL" data-id="/archives">all sites</a> or try a different search.</p>
```
adamwoodnz, line 45: Needs escaping and translation

package.json
```diff
@@ -20,7 +20,9 @@
 		"backstop:test": "npm exec backstop test",
 		"lighthouse": "lighthouse-ci http://localhost:8888/ --accessibility=100 --best-practices=100 --seo=100",
 		"lighthouse:desktop": "lighthouse http://localhost:8888/ --view --preset=desktop --output-path=lighthouse.html",
-		"lighthouse:mobile": "lighthouse http://localhost:8888/ --view --screenEmulation.mobile --output-path=lighthouse.html"
+		"lighthouse:mobile": "lighthouse http://localhost:8888/ --view --screenEmulation.mobile --output-path=lighthouse.html",
+		"lint:theme": "composer run lint source/wp-content/themes/wporg-showcase-2022 && yarn workspace wporg-showcase-2022-theme lint:js && yarn workspace wporg-showcase-2022-theme lint:css",
```
StevenDufresne, line 24: Are these additions intentional? If so, maybe better in a different PR.

StevenDufresne (approved):
I have localized the strings. I'll mention this one more time and if it's still hanging on Monday, we'll merge.

package.json
adamwoodnz: They were, happy to split them out

StevenDufresne:
@adamwoodnz Merge when ready!

PR #87, opened 2022-12-07
Try/configurable archive context tag

### LINKED ISSUE #86: Improve archives heading structure.
We improved the heading structure in https://github.com/WordPress/wporg-showcase-2022/pull/73 but didn't have a clear solution for what to do on the archives page.

StevenDufresne:
Closes #86.

This PR updates the archive context control to make the tag configurable so we can update it to an `<h1>`. This also incidentally updates the template.

Note: This PR borrows some code from the [Gutenberg group block](https://github.com/WordPress/gutenberg/blob/22239e643ca8768ecc76b3261bcc3bcc5eeb1b3e/packages/block-library/src/group/edit.js#L31) to allow users to select the tag that will be used.

### Block Editor Settings
Screen Shot 2022-12-07 at 3 12 38 PM

## How to Test
1. Open the site editor and find the archive template.
2. Select the `archive-results-context` block.
3. Expect to see all settings described in [block.json](https://github.com/WordPress/wporg-showcase-2022/compare/try/configurable-archive-context-tag?expand=1#diff-96670df29bcf4c1991da53ffa3138176044eb73d4d74ae836c7bf18430479005R25)
4. Play with the settings and verify they are applied
5. Change the tag via the new select menu in the advanced section.
6. Save the template, view the template and expect that it uses that tag to render the context string.


adamwoodnz (approved):
Works for me 🚢

PR #92, opened 2022-12-08
Add template mShots generator page

adamwoodnz:


StevenDufresne:
One of the tricky pieces here is that it may be displaying a saved manual screenshot. ([Code Reference](https://github.com/WordPress/wporg-showcase-2022/blob/8bd8059ebfdd85a5ba043ffc0e42370c12149d53/source/wp-content/themes/wporg-showcase-2022/functions.php#L78)). 

**Thoughts:**
- Could we try using different language, `Screenshot Debugger` or `Screenshot Auditor`? 
- Could we identify mshots vs screenshots?
  - Maybe just use CSS to identify which ones are `mShots` and not? Images that use `mShots` will have that in their URL. 
- The block works,  but since it's kind of a utility page does it make sense to add that functionality via a shortcode that returns [edit_post_link](https://developer.wordpress.org/reference/functions/edit_post_link/) to reduce potential debt? I don't know, just a thought.

Lastly, thinking out loud, I'm torn as to whether this should be in the theme itself or just created in the database in the environment. It will be really helpful during the initial launch, not sure if it will be used afterward. hmmm. 🤔  Let's add it to the project... 👍

adamwoodnz:
> it may be displaying a saved manual screenshot

Ah didn't realise that, yep I'll rename and update the copy, and add some differentiation.

> add that functionality via a shortcode

I did try this, but couldn't get a reference to the post id of the current site in the query loop. The shortcode would use the page post id for every item.

> in the theme itself or just created in the database in the environment

Either would be fine, we'd just need a block or working shortcode for the edit link, to be able to build the page.

adamwoodnz:
Updated

localhost_8888_screenshot-auditor_

source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
```diff
@@ -145,3 +145,15 @@
 		}
 	}
 }
+
+.page-template-page-screenshot-auditor {
+	.wp-block-wporg-site-screenshot {
+		img {
+			border: 2px solid var(--wp--preset--color--blueberry-1);
```
adamwoodnz, line 152: I chose the manual upload as the exception, but given how many sites have issues with mShots maybe we want to reverse this?

StevenDufresne:
@adamwoodnz I'm going to merge this. I can use it right now while i'm release planning. 

I think any changes I have requested could've added in the future.

👍

Issue #94, opened 2022-12-08
Update "Browse the archive" link to button with total number of items in archive.


StevenDufresne:
This is a combination of updates in [figma i3](https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=1274%3A21566&t=ntm5rrrhkKGYlBzW-0) and this feedback:

> Design: “Browse the Archive” is harder to find than I would like it to be. We want it to be easy to see that there are more than our 4-5 features sites, and my eyes skipped right over it to “Submit a site”

Source: @josephahaden 

## Screenshot
<img width="626" alt="Screen Shot 2022-12-09 at 8 49 40 AM" src="https://user-images.githubusercontent.com/1657336/206590723-cb1ec01a-6302-470d-9152-e1c2d3043d1d.png">

PR #95, opened 2022-12-09
Add a shortcode to display the number of sites in the archive.

StevenDufresne:
Closes: #94 

This PR adds a shortcode to retrieve the number of published posts to be used within a link on the front page.

| Before | After |
 | --- | --- |
| Screen Shot 2022-12-09 at 10 14 56 AM | Screen Shot 2022-12-09 at 10 13 38 AM |

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/inc/shortcodes.php
+++ b/source/wp-content/themes/wporg-showcase-2022/inc/shortcodes.php
@@ -37,7 +37,7 @@ add_shortcode(
 );
 
 /**
- * Returns the number of published sites in archive.
+ * Returns the number of published posts.
  */
 add_shortcode(
 	'sites_in_archive_count',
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-showcase-2022/inc/shortcodes.php
```diff
@@ -35,3 +35,13 @@ function() {
...
 	}
 );
+
+/**
+ * Returns the number of published posts.
+ */
+add_shortcode(
+	'sites_in_archive_count',
+	function() {
+		return wp_count_posts()->publish;
```
tellyworth, line 45: Pedantically, this should use `_n()` for handling plurals. That's a bit messy to fit with the shortcode method though - you'd probably wind up with the translation string in the shortcode itself. If the count will always be many then it's not worth the bother (though someone will quibble about correctness, there are edge cases for some languages that have varying singular/plural forms for high numbers). I'd consider that not a blocker.

You could use `number_format_i18n()` on the return value.

tellyworth (approved):

PR #96, opened 2022-12-09
Replace the title text "Showcase" with home svg

renintw:
This PR tries to fix the broken breadcrumb appearing on mobile devices. 
The potential problem with this solution is that the layout would still be broken if the title is too long.
However, Haven't seen any so far.
Here's another solution using truncation to fix it.

## Screenshots
Screen Shot 2022-12-09 at 10 35 20 AM


StevenDufresne:
@jasmussen, 

No rush on this one. We're looking for a .org way to fix the breadcrumb when on mobile and the page title is long. 

@renintw Has come up with 2 PRs. 

1. Switch to a home icon on mobile to replace `showcase`
2. Truncate the page when it's too long. (https://github.com/WordPress/wporg-mu-plugins/pull/310)

Any opinions from a design standpoint?

jasmussen:
From those, I'd lean towards keeping the "Showcase" text, and eliding the 2nd level crumb if need be. I.e. 

```
Showcase / The University of Br...        ᐯ
```

We can also make the font size slightly smaller on mobile if need be. And to be clear I'm thinking of this in a generic way across all sections, so we can ideally use the same code/patterns across the site for this.

If we _were_ to use icons, we should use [the new icon set](https://www.figma.com/file/e4tLacmlPuZV47l7901FEs/WordPress-Design-Library?node-id=8%3A13&t=scmBPusZarw1J0gI-1), let's not use Dashicons anywhere on the site:

Screenshot 2022-12-09 at 09 59 38

StevenDufresne:
The breadcrumb will go through more upstreams iterations. We probably don't need this anymore. 

@ren Have any hesitations with closing this PR?

renintw:
> The breadcrumb will go through more upstreams iterations. We probably don't need this anymore. 
> 
> 
> 
> @ren Have any hesitations with closing this PR?

No, I'll close it. 👍

Issue #89, opened 2022-12-07
Accessibility: Showcase hero heading.


StevenDufresne:
> The word 'Showcase' should a) probably not be an image, but definitely b) needs alt text 'Showcase'.

Ref:
https://wordpress.slack.com/archives/C02RP4X03/p1670450092480879?thread_ts=1670382088.221119&cid=C02RP4X03

Props: 
https://profiles.wordpress.org/joedolson/

joedolson:
It's a kind of strange implementation to have a hidden `h1` and a visible SVG with the same text. I didn't even  notice the presence of the H1 on my first look at the page, since it was invisible...

As it is, it's OK for screen reader users; they'll get the hidden heading, and the image will be ignored. But it feels extremely wrong to me. I'd definitely be interested in understanding the reasoning behind that choice; I think it would be quite preferable to find another way.

joedolson:
Yes, right now it's translatable for screen readers only, but not for sighted users.

StevenDufresne:
Yeah, all fair. 

I'll dig up why (I should know I implemented it  -- can't remember) and add a fix with a preference for removing the image. 👍

ryelle:
This pattern of an oversized & cut-off heading in some unique font is/will be common across some of the redesigned sites, so the solution here will probably also be repeated on a few other sites. For example, [these Developers and Jobs mockups](https://make.wordpress.org/design/2022/12/05/design-share-nov-21-dec-2/). IIRC, I recommended the image approach for ease of implementing with Gutenberg, and to avoid adding multiple new fonts to wp.org.

Which is not to say it's the right call - maybe we can add & optimize these fonts like we've done with [the global use fonts](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/global-fonts), and then it wouldn't be as much of a performance issue.

StevenDufresne:
The design team has also requested that the text is indented which will make this more block editor friendly. [Link](https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=1887%3A27708&t=vjFDF4Py6dX4aODg-0#321455685)

![image](https://user-images.githubusercontent.com/1657336/206935448-1795ea91-d4d0-4873-89b7-26ee6accfa85.png)

ryelle:
Header is different in the latest design, closing this issue.

PR #100, opened 2022-12-12
Fix layout to move showcase text on top of site screenshot.

StevenDufresne:
Related #89 

This PR moves the large "Showcase" text to fit within the max width container. Because the text is no longer full-width, I was able to simplify the layout and remove some `CSS`. There should be no visual changes.

This doesn't however, replace the image with the actual text, expected in #89.

Since we are close to launching, I chose to make a smaller change to ensure the hero title does not stop our launch. We don't have the font yet, nor an exact solution on how to achieve the responsive title.

## Screenshots
Note, the "Showcase" image showing in these screenshots is still the one cropped on the left. That will be replaced once these changes are deployed.

| Viewport | Before | After | 
| --- | --- | --- |
 | Mobile | wordpress org_showcase-test_(iPhone 12 Pro) | localhost_8888_(iPhone 12 Pro) (2) | 
| Desktop | wordpress org_showcase-test_ | localhost_8888_ |


adamwoodnz (approved):
Works for me :shipit: 

I kinda liked the cut off 'S' 🤷

PR #102, opened 2022-12-12
Fix/screenshot hover states

### LINKED ISSUE #98: Add hover states for site thumbnails
Could we highlight the thumbnail and the site title on hover? This would be the same approach as in the upcoming [Theme directory redesign](https://www.figma.com/file/DQuNUBmyVbbtJNOqCSrAPH/Theme-Directory?node-id=1819%3A15667&t=qrDWFxUFARIoTYpl-4)

image

StevenDufresne:
Closed #98 

This PR adds hover states to site thumbnails. Previously, we controlled the `border-radius` on a `group` block so editors can update via block editor, however, there is currently no block editor control of `hover` styles so the templates were updated to remove the group and have better control. I also had to add styles for `related posts`.

## Screenshot

Screen Shot 2022-12-12 at 1 01 06 PM

## How to test
You can test the site-screenshot changes locally, however the related post changes need to be tested on a sandbox.


adamwoodnz (commented):
Looking at the design (and what I'd expect), the image and text should both change when hovering over either (basically an invisible card/tile surrounding both). I see we've managed to achieve that for the related sites but not the archive/front page sites. Ideally for accessibility this style would also be applied on focus.

Design

I think the problem here is the markup. Ideally for something like this we'd want a basic card layout like:
```
<ul>
  <li>
    <a href>
      <img />
      <h3>{title}</h3>
    </a>
  </li>
</ul>
```

Which makes styling really easy as we're relying on the natural hover and focus of the anchor element, rather than hovering a non-interactive element like a div.

```
a {
  img {
    border: ...
  }

  &:hover,
  &:focus {
    img {
      border: ...
    }

    h3 {
      color: ...
      text-decoration: ...
    }
  }
}
```

Is it possible to wrap elements in anchors via the editor? Maybe we've hit the limits in terms of creating high quality interactive layout...

adamwoodnz:
This might do for launch anyway, and maybe we can iterate to make it more accessible

adamwoodnz:
Sorry, I can see this on focus actually, but not both elements a once. Also the related title gets focus before the image.

StevenDufresne:
> > Looking at the design (and what I'd expect), the image and text should both change when hovering over either (basically an invisible card/tile surrounding both). I see we've managed to achieve that for the related sites but not the archive/front page sites. Ideally for accessibility this style would also be applied on focus.
> 
> I'm assuming by this you mean the when hovering over the title, the image doesn't get the outline?
> 
> > I think the problem here is the markup. Ideally for something like this we'd want a basic card layout like:
> 
> Yeah, the markup is definitely the issue. We could remove the `isLink:true` from the site-screenshot & from the post title and then wrap the site-screenshot and the title in `a` tag (but would also require a `p` tag through the editor). Otherwise we would need to compose our own component which I'm not keen on doing.
> 
> I'll give it a try.

Actually, no you can't wrap anything in a `<a>` via the editor.

adamwoodnz:
> Actually, no you can't wrap anything in a <a> via the editor.

Something to feed back upstream I think. It's a pretty common pattern.

StevenDufresne:
> > Actually, no you can't wrap anything in a  via the editor.
> 
> Something to feed back upstream I think. It's a pretty common pattern.

I'll spin up a wrapper component.

adamwoodnz:
> when hovering over the title, the image doesn't get the outline?

Yeah hovering the title doesn't highlight the image, and vice versa

StevenDufresne:
Closing in favor of #104

Issue #109, opened 2022-12-14
Move the URL of the site under the title


javierarce:
The site's URL appeared originally at the end of each article, but we have been asked to move it right under the title:

![image](https://user-images.githubusercontent.com/4933/207814622-c526afdd-2355-4936-8839-cfca55d21031.png)

Updated: added new screenshot with the external link icon

PR #110, opened 2022-12-15
Move site link under title.

StevenDufresne:
Fixes: #109.

This PR moves the site URL under the post-title. I have opened this as a PR because the figma designs do not include the external link icon but it was made a .org requirement. I want to surface that before this change merges. 

| Design | Implementation |
 |--- | --- |
| Screen Shot 2022-12-15 at 9 03 07 AM | Screen Shot 2022-12-15 at 9 06 24 AM |

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/patterns/page-single.php
+++ b/source/wp-content/themes/wporg-showcase-2022/patterns/page-single.php
@@ -7,14 +7,12 @@
 
 ?>
 
-<!-- wp:post-title {"level":1,"style":{"spacing":{"margin":{"bottom":"var:preset|spacing|30"}}},"fontSize":"heading-2"} /-->
-
-<!-- wp:post-content {"align":"full","layout":{"type":"constrained"}} /-->
-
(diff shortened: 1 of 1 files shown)


javierarce:
Sorry, not including the icon in the design was an oversight 🚀

javierarce:
Just one change: we are using the symbol ↗︎ instead of the SVG icon

image

StevenDufresne:
@javierarce I mentioned the issue in https://github.com/WordPress/wporg-main-2022/pull/157#discussion_r1049231420.

We seem to not be consistently using that arrow. On some pages, it points to external links and on some pages, it doesn't. I think we need to find some consistency here. What is it denoting? If it's denoting just a link (like it is on wordpress.org/gutenberg, or the state of the word banner on that homepage) then it becomes unclear which ones are external. In my personal, non-designer, opinion I don't know what those arrows add to links. Links in themselves denote navigation, I'm not certain why we need them for internal links as an accent.

javierarce:
I'm sorry, @StevenDufresne, you are 100% right. I was confused by our current use of the diagonal icon in other places (we will revisit those decisions). 

Let's go with the classic external icon but set in blueberry (#3858E9)

Issue #80, opened 2022-12-05
Hide Submit a Site link in the navigation of the submission form


javierarce:
The Submit a Site page shouldn't link to itself from the navigation bar. I've updated [the design](https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=1%3A2193) to reflect that.

![image](https://user-images.githubusercontent.com/4933/205663429-6616a298-8c15-49c6-9814-cffef7313cc8.png)

StevenDufresne:
I think this will be trickier than it seems. Can we push this to a future iteration?

javierarce:
Yes, of course, I think this can wait.

PR #111, opened 2022-12-15
Turn off submit site link on that page.

StevenDufresne:
Fixes: #80 

This PR hides the "Submit a site" link when viewing that page. I create a new template part to accomplish this. The con of this approach is that changes to the subnav styles will need to be made in 2 places now. However, I don't think this is a big risk and is less than likely.

**Alternative Approaches**
- I looked at targeting via CSS but because there is some mobile code that adds a dropdown button, we would get an empty menu. If we only hide it on desktop, it's a bit weird to have different content.
- I briefly considered adding some php/js logic but it seems like overkill.

## How To Test
- Visit `/submit-a-wordpress-site`, expect to have no "Submit a site" navigation link.

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/parts/page-submit-header.html
+++ b/source/wp-content/themes/wporg-showcase-2022/parts/page-submit-header.html
@@ -1,3 +1,5 @@
+<!-- wp:wporg/global-header {"style":"black-on-white"} /-->
+
 <!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|charcoal-1"}}},"spacing":{"padding":{"top":"16px","right":"var:preset|spacing|60","bottom":"16px","left":"var:preset|spacing|60"}},"border":{"top":{"color":"var:preset|color|light-grey-1","width":"1px","style":"solid"}}},"backgroundColor":"white","textColor":"charcoal-1","className":"is-sticky","layout":{"type":"constrained"}} -->
 <div class="wp-block-group alignfull has-charcoal-1-color has-white-background-color has-text-color has-background has-link-color is-sticky" style="border-top-color:var(--wp--preset--color--light-grey-1);border-top-style:solid;border-top-width:1px;padding-top:16px;padding-right:var(--wp--preset--spacing--60);padding-bottom:16px;padding-left:var(--wp--preset--spacing--60)"><!-- wp:group {"align":"full","layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between"}} -->
 <div class="wp-block-group alignfull"><!-- wp:wporg/site-breadcrumbs {"textColor":"charcoal-1","fontSize":"small"} /-->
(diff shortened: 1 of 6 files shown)


adamwoodnz (commented):
This works for me. 

I find the structure of the header a bit confusing though. Is `patterns/subnav-bar.php` intended to be a reusable pattern, or could that markup just be inline in `parts/header.html`?

And if so I wonder if it might be simpler to have a second `parts/page-submit-header.html` which is very similar but without the navigation menu:

header.html
```
<!-- wp:wporg/global-header {"style":"black-on-white"} /-->

<!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|charcoal-1"}}},"spacing":{"padding":{"top":"16px","right":"var:preset|spacing|60","bottom":"16px","left":"var:preset|spacing|60"}},"border":{"top":{"color":"var:preset|color|light-grey-1","width":"1px","style":"solid"}}},"backgroundColor":"white","textColor":"charcoal-1","className":"is-sticky","layout":{"type":"constrained"}} -->
<div class="wp-block-group alignfull has-charcoal-1-color has-white-background-color has-text-color has-background has-link-color is-sticky" style="border-top-color:var(--wp--preset--color--light-grey-1);border-top-style:solid;border-top-width:1px;padding-top:16px;padding-right:var(--wp--preset--spacing--60);padding-bottom:16px;padding-left:var(--wp--preset--spacing--60)"><!-- wp:group {"align":"full","layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between"}} -->
<div class="wp-block-group alignfull"><!-- wp:wporg/site-breadcrumbs {"textColor":"charcoal-1","fontSize":"small"} /-->


<!-- wp:navigation {"ref":7750,"textColor":"blueberry-1","backgroundColor":"white","className":"is-style-dropdown-on-mobile","style":{"spacing":{"blockGap":"24px"}},"fontSize":"small"} /--></div>
<!-- /wp:group --></div>
<!-- /wp:group -->

```

page-submit-header.html
```
<!-- wp:wporg/global-header {"style":"black-on-white"} /-->

<!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|charcoal-1"}}},"spacing":{"padding":{"top":"16px","right":"var:preset|spacing|60","bottom":"16px","left":"var:preset|spacing|60"}},"border":{"top":{"color":"var:preset|color|light-grey-1","width":"1px","style":"solid"}}},"backgroundColor":"white","textColor":"charcoal-1","className":"is-sticky","layout":{"type":"constrained"}} -->
<div class="wp-block-group alignfull has-charcoal-1-color has-white-background-color has-text-color has-background has-link-color is-sticky" style="border-top-color:var(--wp--preset--color--light-grey-1);border-top-style:solid;border-top-width:1px;padding-top:16px;padding-right:var(--wp--preset--spacing--60);padding-bottom:16px;padding-left:var(--wp--preset--spacing--60)"><!-- wp:group {"align":"full","layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between"}} -->
<div class="wp-block-group alignfull"><!-- wp:wporg/site-breadcrumbs {"textColor":"charcoal-1","fontSize":"small"} /-->

</div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```

Then there would only be one file switched between the templates.

StevenDufresne:
> This works for me.
> 
> I find the structure of the header a bit confusing though. Is `patterns/subnav-bar.php` intended to be a reusable pattern, or could that markup just be inline in `parts/header.html`?

It originally had a translatable string in it but the site-breadcrumbs PR removed it. It can now be within a template part. I think that should happen in a separate PR though.


> And if so I wonder if it might be simpler to have a second `parts/page-submit-header.html` which is very similar but without the navigation menu:

Yep, that's a good idea. I'll create a new header.

adamwoodnz (approved):
LGTM :shipit:

PR #121, opened 2023-01-02
Restrict font size on grid items on mobile.

### LINKED ISSUE #114: Reduce the font size of site's titles in mobile
In small viewports, the site titles should be set in smaller font size. They are currently 20px, but they should appear in 16px

Before/After image: 
image

StevenDufresne:
Fixes #114 

This PR updates the post-title to be`16px` on mobile.

**Considerations**
- The theme currently uses `em` units for `h*` margins. On mobile, the space between the top of the heading and the image could be considered tight. Leaving for now unless otherwise advised to implement a standard margin.
- `!important` is needed because theme.json styles also have that declaration.

### Homepage
| Before | After |
| --- | --- |
|  localhost_8888_(iPhone 12 Pro) (1) | localhost_8888_(iPhone 12 Pro) | 

### Search Results
| Before | After |
| --- | --- |
|  localhost_8888__s=c(iPhone 12 Pro) (1) |  localhost_8888__s=c(iPhone 12 Pro) |

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
+++ b/source/wp-content/themes/wporg-showcase-2022/src/style/style.scss
@@ -47,6 +47,16 @@
 }
 
 /*
+ * core/post-title.
+ */
+
+@media (max-width: 600px) {
+	.wp-block-post .wp-block-post-title {
+		font-size: 16px !important;
+	}
+}
+
+/*
  * Related Posts Styles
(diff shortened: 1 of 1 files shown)


adamwoodnz (commented):
LGTM if we only want post title to change and not everything `large`, as noted in @ryelle's comment linked above

PR #125, opened 2023-06-22
Initial pass at next showcase iteration

ryelle:
I started poking at the theme while reading through the latest design post, and before I knew it I'd made quite a few changes wireframing out the updates. I don't know if we want to merge this with the placeholder content, or if it would be better to split it apart (I mean, probably, but it was tricky enough splitting up into different commits), but I thought I'd make the PR either way.

The commits might not individually work, but step by step they explain the changes. Mostly it's trying to DRY the patterns & templates, fixing up some block spacing, and removing unused files.

| Home | Single site |
|---|---|
| sc-home | sc-single |
| Search | Tag archive |
| sc-search | sc-tag |
| Home (small) | Submit site (small) |
| sc-home-small | sc-submit-small |


source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/style.scss
```diff
@@ -1,22 +1,24 @@
...
+
 	img {
 		width: 100%;
 		vertical-align: middle;
 		aspect-ratio: 16 / 9;
 	}
 }
-
-// If we have a parent block with a border radius, we want to clip the contents.
-.wp-block-group[style*="border-radius"]:has(.wp-block-wporg-site-screenshot) {
```
StevenDufresne, line 10: These rules were to allow for editor customization of the border-radius. We may want to keep them.

StevenDufresne:
Since most of this is templating, I personally feel comfortable with merges directly to trunk. :thumbs

source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/style.scss
ryelle: We don't need it now that the block itself supports the border radius.

PR #126, opened 2023-07-05
Add support for two screenshots per site

ryelle:
The design currently has two screenshots in the single page header, a desktop and mobile screenshot. The archive grid templates also use a screenshot with the same aspect ratio as the desktop image. So we can use two images per site to achieve this look, one desktop & one mobile.

This PR updates the current screenshot preview block (generated using mshots) and adds new meta fields for custom uploads, along with a UI in the editor.

The expectation is that the screenshots will be resized to display at 700 × 370 & 170 × 300 on the single page, so those are the aspect ratios used for capturing from mshots. Specifically, we capture at a viewport of 1440px × 761px & 480px × 847px — those widths can be adjusted, but the height is generated from the display size.

I've switched the `wporg/site-screenshot` to use a `type` attribute instead of `useHiRes`, thinking that we might need to add additional types/sizes later.

| Editor | Single | Grid |
|---|---|---|
| editor | showcase-header | Screen Shot 2023-07-05 at 14 13 25 |

If there's an issue uploading, a notice is displayed above the upload drop area.
Screenshot 2023-07-05 at 1 56 00 PM

<details>
<summary>Audit page: long screenshot, can see both custom (blue borders) and generated images.</summary>

screenshot-audits

</details>

To test
1. Build the branch
2. Generated screenshots
    - View a site (maybe reload, so mshots can work)
    - The screenshots should be from `https://wordpress.com/mshots/v1/…`
    - There should be two screenshots, one visibly a desktop size and one mobile size
    - They should not be distorted on either the homepage grid or single pages
3. Custom screenshots
    - Edit a site, see the Screenshots panel in the document sidebar
    - Add some images
    - They should appear on the frontend of the site
    - They should not be distorted on either the homepage grid or single pages

No design review needed yet, the design is not final so all I've really done is dropped the blocks on the page.

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
@@ -47,10 +47,6 @@ function render( $attributes, $content, $block ) {
 
 	$screenshot = site_screenshot_src( $post );
 
-	if ( isset( $attributes['useHiRes'] ) && true === $attributes['useHiRes'] ) {
-		$screenshot = add_query_arg( 'scale', 2, $screenshot );
-	}
-
 	$loading = 'eager';
 	if ( isset( $attributes['lazyLoad'] ) && true === $attributes['lazyLoad'] ) {
 		$loading = 'lazy';
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
@@ -8,7 +8,7 @@
 
 namespace WordPressdotorg\Theme\Showcase_2022\Site_Screenshot;
 
-use function WordPressdotorg\Theme\Showcase_2022\site_screenshot_src;
+use function WordPressdotorg\Theme\Showcase_2022\get_site_domain;
 
 add_action( 'init', __NAMESPACE__ . '\init' );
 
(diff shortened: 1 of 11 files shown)


source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
```diff
@@ -45,18 +45,19 @@ function render( $attributes, $content, $block ) {
...
 	$loading = 'eager';
 	if ( isset( $attributes['lazyLoad'] ) && true === $attributes['lazyLoad'] ) {
 		$loading = 'lazy';
 	}
 
-	$img_content = "<img src='" . esc_url( $screenshot ) . "' alt='" . the_title_attribute( array( 'echo' => false ) ) . "' loading='" . esc_attr( $loading ) . "' />";
+	$img_content = sprintf(
+		'<img src="%1$s" alt="%2$s" loading="%3$s" />',
+		esc_url( $screenshot ),
+		the_title_attribute( array( 'echo' => false ) ),
```
StevenDufresne, line 58: Can we change this to use the image alt text?

StevenDufresne (approved):
Nice! Way better interface!

A couple minor points.

source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
ryelle: This image could be an mshots image, and we don't really know anything about it at this point since `get_site_screenshot_src` only returns a URL. But I do want to revisit how these are linked/alt-text'd in the future.

ryelle: (I'm going to hold off on creating an issue for that until we have the final design and know how these screenshots will be used)

PR #127, opened 2023-07-05
Register meta fields and add a panel to the editor to manage the content

ryelle:
Create a panel in the document sidebar to handle the custom meta fields: domain, country, author, and theme. These are plain text inputs, no validation is done (except for the default schema validation).

editor

We can iterate on the fields, types, and help text.

Only domain is used currently, but the rest should save when edited. These will be used in the "more about" section, if they exist.

Screenshot 2023-07-05 at 2 35 39 PM

**To test**
- Edit a site
- There may be values from the existing site data, these are the same meta keys
- Update some values
- Save the post, it should save
- Reload the page and/or check the domain on the frontend, it should be the content you just entered


source/wp-content/themes/wporg-showcase-2022/src/meta-box/index.js
```diff
@@ -0,0 +1,62 @@
...
+			/>
+			<TextControl
+				label={ __( 'Country', 'wporg' ) }
+				help={ __( 'Location of site.', 'wporg' ) }
+				onChange={ onChange( 'country' ) }
+				value={ meta.country }
+			/>
+			<TextControl
+				label={ __( 'Author', 'wporg' ) }
+				help={ __( 'Author or agency name.', 'wporg' ) }
```
StevenDufresne, line 46:
```suggestion
				help={ __( 'Author or agency name, if known.', 'wporg' ) }
```

StevenDufresne (approved):
Looks good to me.

PR #149, opened 2023-09-13
Add related sites to single site pages

### LINKED ISSUE #140: Related sites on single sites
Related sites should show up below the pagination at the end of the single site view. See #132.

This is a grid of three sites, and it does not drop to two-column, the cards are allowed to shrink more.

Related

If there are no related sites, grab some recent sites. See [customize related posts](https://jetpack.com/support/related-posts/customize-related-posts/) for some helper hooks.

ryelle:
Fixes #140 — This adds the related sites section to single site pages. It uses Jetpack's related posts feature, but overrides the output to render using the theme's blocks. It also adds some fallback posts to make sure there are always 3 results in the related section.

| Design | Branch |
|---|---|
|  |  |
|  |  |

@WordPress/meta-design The heading font size (for "Related sites") uses the "heading 5" font size, which is 26px on large screens, but breaks to 20px on small. In figma, it's 26 across both sizes, is that intentional? Should I set this to a custom 26px instead?

**To test:**

This needs to be tested on a live site, like the staging site, so use your sandbox. This is because Jetpack's related posts does not work in dev/offline mode.

- View a single site
- It should have 3 posts below the pagination
- Sites without related content will use the latest 3 featured sites
- Older sites will have real related content, for example https://wordpress.org/showcase-v2/the-city-university-of-new-york/
- Sites with less than 3 related posts will fill in with featured sites


StevenDufresne (approved):
Tested on my sandbox. Looks good!

jasmussen:
Nice, very close. There's a small thing with the text and tags not lining up:
267767574-e27151e8-48c1-4641-8994-65b6de1919fb

It's probably a line-height that is affected by whether tags are present or not.

ryelle:
Alignment updated:

Screenshot 2023-09-14 at 1 38 57 PM

PR #172, opened 2023-09-21
Site screenshot: Load mshots images via JS, show CSS loading state

### LINKED ISSUE #154: Improve mShots mobile loading window.
When mShots is creating a new snap shot it returns a "Generating Preview" image with dimensions that don't match the final output of the preview window so the mobile phone snaps upwards when an image loads. Can we maybe replace that image with a css loading implementation? 

That will give us:
1. A nicer loading state
2. Remove the upwards jump.

## Image
Screenshot 2023-09-18 at 6 55 44 AM

ryelle:
Fixes #154 — I've used the basics from the Screenshot Preview block, but I didn't want to load all of React for a single component. I tried the [Interactivity API](https://github.com/WordPress/gutenberg/blob/trunk/packages/interactivity/docs/2-api-reference.md) again and I think it's worked fairly well. It follows the same behavior — the default state is "loading" with a CSS spinner. The `init` function is called by the `wp-init` directive when the node is created, and that triggers the initial fetch and subsequent fetches. Once the image is not redirecting, we assume it's valid and update the context. The `update` function is called on the `wp-effect` directive, so it's called each time the context changes. This function controls the actual display of the image — once we have a `base64Image` it's loaded in here. This also handles the error state, simply displaying the alt text (site title) — hopefully this isn't needed often :)

There are still awkward cases when the site URL is invalid, because mshots does not communicate that back on the response. but those are cases we would want to intentionally handle, either removing the site or uploading a proper image, so maybe that's fine?

This does introduce a new design element, the spinner. I'm using the same color as the border to keep it unobtrusive.

https://github.com/WordPress/wporg-showcase-2022/assets/541093/4de316e3-fcdc-401d-8532-8b37eda07f64

I had it fully regenerate the images here, in most cases the images would load faster than this. On the plus side, you can see the spinner design :)

https://github.com/WordPress/wporg-showcase-2022/assets/541093/71f0bb67-e4fe-4e20-aa8c-e592a3de68da

This also removes [the aspect ratio added here](https://github.com/WordPress/wporg-showcase-2022/commit/abf3414a3d55306b26b101d9da8605b291646892), because it was squishing some of the manually updated images.

**To test**

1. Check out this branch, build the theme
2. Load any page - archives or single
3. Sites that have local images should have no changs
4. Sites without uploaded screenshots will now use the CSS loader rather than the mshots loading gif
5. mshots images should still load correctly

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/block.json
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/block.json
@@ -1,6 +1,6 @@
 {
 	"$schema": "https://schemas.wp.org/trunk/block.json",
-	"apiVersion": 2,
+	"apiVersion": 3,
 	"name": "wporg/site-screenshot",
 	"version": "0.1.0",
 	"title": "Site Screenshot",
@@ -54,5 +54,7 @@
 	"usesContext": [ "postId" ],
 	"editorScript": "file:./index.js",
 	"editorStyle": "file:./index.css",
-	"style": "file:./style-index.css"
+	"viewScript": "file:./view.js",
+	"style": "file:./style-index.css",
+	"render": "file:../../src/site-screenshot/render.php"
 }
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/index.php
@@ -22,57 +22,7 @@ add_filter( 'render_block_core/group', __NAMESPACE__ . '\inject_background_color
  * @see https://developer.wordpress.org/reference/functions/register_block_type/
  */
 function init() {
-	register_block_type(
-		dirname( dirname( __DIR__ ) ) . '/build/site-screenshot',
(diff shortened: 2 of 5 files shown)


StevenDufresne (approved):
So much better!

jasmussen:
Vastly better.

My only request, which should not be blocking, would be to use either https://wordpress.github.io/gutenberg/?path=/docs/components-spinner--docs or https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-progressbar--docs instead. 

If using those components is too tricky, you can use this lightweight version: https://codepen.io/joen/pen/yLGvLeO?editors=1100

ryelle:
Updated the spinner:

Screenshot 2023-09-22 at 11 34 37 AM

jasmussen:
One thing to follow up on, not urgent: the spinner should be 20x20px. Intentionally small!

ryelle:
Can you create a new issue for that?

StevenDufresne:
Done: https://github.com/WordPress/wporg-showcase-2022/issues/177

PR #175, opened 2023-09-22
Site screenshot: Add responsive image support to the screenshot block

### LINKED ISSUE #156: Images seem larger than they need to be.
The image file sizes appear too large and result in a heavy page load burden (which affects performances scores).

If this is solely the result of missing screenshots, feel free to close this.

We should: 
- Make sure images are compressed before upload (or on upload)?
- Provide more appropriately sized images

## Screenshots
Screenshot 2023-09-18 at 9 29 03 AM
Screenshot 2023-09-18 at 9 32 23 AM

### LINKED ISSUE #171: Breakpoint change
Live version has 3 columns after a breakpoint (1440px). We discussed having that breakpoint at 1920px, as in design.

1. What bugs can happen if we do that?
2. Is it lot of work to do it?

Screen Shot 2023-09-21 at 15 31 00

CC: @jasmussen @ryelle

ryelle:
Fixes #156. This adds a new attribute to the screenshot block, `location`. This passes a hint to the block about where the image will be used, so that we can set location-aware responsive image hints. When location is `grid`|`row` (related sites), the block now adds `srcset/sizes` attributes.

This PR also sets up a few new image sizes from 500-1400px wide, which covers the different sizes the images are shown in the grid. Locally you can regenerate these with `wp media regenerate`, but sandboxes & production will use Photon and dynamically load the right size.

Using smaller (height/width) images decreases the filesizes, so most images are under 500KB now even on retina screens.

This also fixes #171, because changing that later would require changing the `sizes` value too.

**To test**

- View anywhere the site screenshot is displayed, there should be no visible difference and no image quality change.
- Images in the site grid should have `srcset` and `sizes` attributes
- If you watch the network tab, the image loaded should reflect the column size (and screen resolution).

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/functions.php
+++ b/source/wp-content/themes/wporg-showcase-2022/functions.php
@@ -62,6 +62,17 @@ function setup_theme() {
 	add_image_size( 'screenshot-desktop', 2044, 1150, array( 'center', 'top' ) );
 	add_image_size( 'screenshot-mobile', 750, 1334, array( 'center', 'top' ) );
 
+	// More desktop sizes to support responsive images in the grid.
+	// In production these will be generated using Photon CDN, but we can set
+	// them up as real sizes for local support too.
+	// Images in the grid are display from 340-740px wide. This set of sizes
+	// should cover 1x and 2x resolution.
+	$desktop_ratio = 1150 / 2044;
+	add_image_size( 'screenshot-desktop-1400', 1400, 1400 * $desktop_ratio, array( 'center', 'top' ) );
+	add_image_size( 'screenshot-desktop-1100', 1100, 1100 * $desktop_ratio, array( 'center', 'top' ) );
+	add_image_size( 'screenshot-desktop-800', 800, 800 * $desktop_ratio, array( 'center', 'top' ) );
+	add_image_size( 'screenshot-desktop-500', 500, 500 * $desktop_ratio, array( 'center', 'top' ) );
+
 	// Add tonesque support so that Jetpack loads the class.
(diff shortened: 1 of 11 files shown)


adamwoodnz (commented):
👍 Working for me in the grid.

Is there a reason not to do it on the single site page? At mobile size it looks like we're still loading the largest size:

Screenshot 2023-09-27 at 3 08 25 PM

> sandboxes & production will use Photon and dynamically load the right size

Want to make sure I understand this, does photon look at the `srcset` and `sizes` and set the src appropriately?

StevenDufresne (approved):
Great improvement.

renintw (approved):
Look good to me 👍 I can see the srcset and sizes in grid and row.

image

ryelle:
> Is there a reason not to do it on the single site page? At mobile size it looks like we're still loading the largest size:

No, I just focused on grid pages first, having the most images and therefore highest impact. It could be updated to work on single sites in the future, all that's needed is to figure out what the sizes & srcsets should be.

> Want to make sure I understand this, does photon look at the srcset and sizes and set the src appropriately?

No, this is just about generating those image URLs. Photon hooks into the way WP generates media URLs and replaces them, so when you call for an image that's 100x100, it will automatically add `resize=100,100` to the returned URL. So it doesn't matter if the local site has generated that image, the CDN will resize automatically.

StevenDufresne:
I was navigating around the site on staging and found the single pages were pretty slow to load. 

Can we update them? The related sites contribute to the slowness.

Page: https://wordpress.org/showcase-v2/opus-one/ 
Screenshot 2023-10-04 at 12 09 10 PM

PR #217, opened 2023-10-02
Add breadcrumbs to all pages

### LINKED ISSUE #208: Add breadcrumbs to all pages
Here are the breadcrumbs, and they should be found on 'search results' as well...

Taxonomy for search results: "Home / Showcase / Search results"

image

CC: @jasmussen

ryelle:
Fixes #208 — this adds the Site Breadcrumbs block to all pages except the homepage. It also customizes the output of the block to fit the hierarchy of Showcase. Most pages use the default white background, but single posts (sites) and the submission "Thanks" page use a dark bar.

**Screenshots**

Ignore the site content on my staging site, I haven't updated it since we started. The only relevant change here is the breadcrumb.

No change to home

b-home

Archives

| All sites | Single filter | Multiple filters | Search results |
|---|---|---|---|
| b-archives | b-archive-single-filter | b-archive-multiple-filters | b-archive-search |

Singles, note that single sites use a dark background.

| Site | Page (tags page) |
|---|---|
| b-single | b-filters |

Submission

| Log in | Submission form | Thanks |
|---|---|---|
| b-submit-login | b-submit | b-submit-thanks |

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/inc/block-config.php
+++ b/source/wp-content/themes/wporg-showcase-2022/inc/block-config.php
@@ -371,10 +371,11 @@ function update_site_breadcrumbs( $breadcrumbs ) {
 		return $breadcrumbs;
 	}
 
-	$term_names = get_applied_filter_list( false );
-
-	 // This matches the "posts page", the All Sites page.
+	// `is_home` matches the "posts page", the All Sites page.
+	// `is_archive` matches any core archive (category, date, etc).
 	if ( is_home() || is_archive() ) {
+		// Get the current applied filters (except search, handled above).
+		$term_names = get_applied_filter_list( false );
 		if ( empty( $term_names ) ) {
 			$breadcrumbs[] = array(
 				'url' => false,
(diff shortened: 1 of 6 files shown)


StevenDufresne (commented):
This works well!

I personally don't like when elements appear to be connected but are not. For example, the single page breadcrumb looks like it's part of the subnav but on scroll, it's revealed to be separate. It's definitely not a deal breaker, but the interface should be as predictable as possible.

Screenshot 2023-10-03 at 12 16 53 PM

ryelle:
> I personally don't like when elements appear to be connected but are not. For example, the single page breadcrumb looks like it's part of the subnav but on scroll, it's revealed to be separate.

yeah especially there it looks weird. https://github.com/WordPress/wporg-mu-plugins/pull/466 + #218 would fix that though.

jasmussen:
I personally think this can work, but I'm going to share it with Marko for a sanity check. 

I think especially the "thank you" page either needs to have this PR in place, _or_ the page itself needs to be light.

marko-srb (changes requested):
Yes, all works except the 'Single page'. Can we revert white/light breadcrumb there? 

Thank you!

jasmussen:
Happy to go with Marko's suggestion 👍 👍

marko-srb:
Tags page. 

"Filters" breadcrumb may be "Wayfinding"? If doesn't seem right, just leave it as filters...

PR #228, opened 2023-10-08
Query Filters: Add taxonomy term counts

### LINKED ISSUE #225: Add taxonomy counts to the filter dropdowns
User feedback indicated that the "Apply" button was a bit confusing. The fact that it includes "(x)", the user thought that was how many sites were included in the selected taxonomy. It was not clear that this number was how many taxonomy terms were selected. 

Perhaps one solution is to add site counts for each term. It makes the filter dropdown more cluttered, but perhaps it provides more context to the user. 

image

cc @jasmussen @marko-srb

renintw:
Closes #225 

This PR adds taxonomy term counts to the filter dropdowns.

Screen Shot 2023-10-09 at 3 59 42 AM

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/inc/block-config.php
+++ b/source/wp-content/themes/wporg-showcase-2022/inc/block-config.php
@@ -70,13 +70,7 @@ function get_post_tag_options( $options ) {
 		_n( 'Popular tags <span>%s</span>', 'Popular tags <span>%s</span>', $count, 'wporg' ),
 		$count
 	);
-	$option_labels = array_map(
-		function( $name, $count ) {
-			return $name . " ($count)";
-		},
-		wp_list_pluck( $tags, 'name' ),
-		wp_list_pluck( $tags, 'count' )
-	);
+	$option_labels = get_option_labels( $tags );
 	return array(
 		'label' => $label,
 		'title' => __( 'Popular tags', 'wporg' ),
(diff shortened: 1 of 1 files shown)


adamwoodnz (commented):
Nice, works on categories for me.

I was expecting numbers on the other filters (tags and flavors) too though. I think the same confusion would exist for them. Any reason not to?

renintw:
Ah, yeah, you are right. I didn't notice that each taxonomy has its own function. Thanks!

**Updated Screenshots**

Screen Shot 2023-10-09 at 12 06 32 PM
Screen Shot 2023-10-09 at 12 06 36 PM

source/wp-content/themes/wporg-showcase-2022/inc/block-config.php
```diff
@@ -70,12 +70,19 @@ function ( $a, $b ) {
 		_n( 'Popular tags <span>%s</span>', 'Popular tags <span>%s</span>', $count, 'wporg' ),
 		$count
 	);
+	$option_labels = array_map(
+		function( $name, $count ) {
+			return $name . " ($count)";
+		},
+		wp_list_pluck( $tags, 'name' ),
+		wp_list_pluck( $tags, 'count' )
+	);
```
adamwoodnz, line 79: How about adding a local utility function to reuse for all these?

renintw: Added in 468d05315ea3096cd25e30c1e9006403fc0169d2

source/wp-content/themes/wporg-showcase-2022/inc/block-config.php
```diff
@@ -421,3 +403,22 @@ function update_site_breadcrumbs( $breadcrumbs ) {
 
 	return $breadcrumbs;
 }
+
+/**
+ * Get an array of option labels.
+ * Each label is a string formatted as 'name (count)'. e.g, Business (30).
```
adamwoodnz, line 409: ❤️

adamwoodnz (approved):
LGTM! :shipit:

Issue #183, opened 2023-09-27
Switch tags with categories in site query loops


ndiego:
Given the relatively small about of sites in the planned initial launch of Showcase v2, there are only a few sites for each tag. The tags located under each site in the query loops are compelling to click on, but this often leads you to a lonely page of 1-2 sites. 

Displaying the site categories instead of tags will lead to a better user experience. Since there are only a handful of categories, most have 10+ results. 

<img width="800" alt="image" src="https://github.com/WordPress/wporg-showcase-2022/assets/4832319/0e143bc0-ed61-4d98-94f1-93fee3a8a177">

ryelle:
It should also hide the Featured category on the homepage, since all items there will have that category, right?

ndiego:
Yup, that would be great, thanks!

StevenDufresne:
Apologies for the late response, I missed this issue. I'll add my thoughts so they can represented. I don't think we need to revert, but I think they are worth noting.

I find this switch takes away from the vibrancy of the directory. The tags create depth in the tiles, and words like "Grafitti, Arts, Design, Culture" are intriguing. Swapping that out for "Creative" leaves more to be desired. 

A lot of the tiles have the word "Business". I couldn't imagine a more uninspiring link to click on.

We've discussed how important it is for users to imagine themselves building something inspiring to be displayed in the showcase and the submit flow lighting the flame. We shouldn't shy away from sparse archives but embrace the quietness and use it to engage the viewer.

If we have less than `x` entries for a specific tag, we could add a skeleton frame or section that says something to the effect:

```
We don't have many themes for `{tag}`, know of any good ones? Submit a cool {tag} theme for the Showcase.

No? Ready to start building your own?  [flow to download/try wp-playground, etc...]
```
We could also add a new section that displays a semi-random selection to fill in the space (we already do that in the [pattern directory](https://wordpress.org/patterns/search/234sdfasdfasdf/)).

```
More showcase sites you may be interested in:

[semi-random selection]
```

StevenDufresne:
To me, the second version does more to communicate how WordPress is being used in the wild.

<img width="1389" alt="Screenshot 2023-10-10 at 12 43 16 PM" src="https://github.com/WordPress/wporg-showcase-2022/assets/1657336/431ccd2e-c7c4-4271-a130-a0803170b3cc">
[...]

PR #232, opened 2023-10-09
Update the Post Terms block to display categories in the Jetpack Related Posts block.

ndiego:
The Jetpack Related Posts block currently displays site tags. Since we [switched to displaying categories](https://github.com/WordPress/wporg-showcase-2022/issues/183) in Query loops, it makes sense to update the Related Sites section to display categories as well.


StevenDufresne:
This works as expected. 👍 

Apologies for the late take, I missed the original change. I'll respond there with my thoughts.

StevenDufresne (approved):

ndiego:
I am going to merge now since everything is using categories. But if we ever want to switch back to tags, it's an easy switch.

Issue #233, opened 2023-10-10
Lighthouse audit


adamwoodnz:
Report generated using pagespeed.web.dev https://pagespeed.web.dev/analysis/https-wordpress-org-showcase-v2/ohj0httcll?form_factor=desktop

### Desktop

![Screenshot 2023-10-10 at 5 19 07 PM](https://github.com/WordPress/wporg-showcase-2022/assets/1017872/4b6622b2-99f4-4613-8303-8cbb82e02246)

### Mobile

![Screenshot 2023-10-10 at 5 20 48 PM](https://github.com/WordPress/wporg-showcase-2022/assets/1017872/e530d813-4019-4735-bbc0-433c9cf613f7)

## Findings

✅ Best Practices and SEO

The desktop accessibility issue is related to the global header aria roles so is not directly related to this effort.

The performance scores are impacted the most by the [Largest Contentful Paint metric](https://developer.chrome.com/docs/lighthouse/performance/lighthouse-largest-contentful-paint).

### Ways we can improve this:

![Screenshot 2023-10-10 at 5 24 59 PM](https://github.com/WordPress/wporg-showcase-2022/assets/1017872/0e699a83-dbad-41e9-aba0-1e7147f7ea76)

#### Eliminating render blocking resources

- [x] `wporg-two-factor` styles [should not be loaded](https://github.com/WordPress/wporg-two-factor/pull/255)
- ~[ ] Enqueue any other styles and scripts in the footer or with `defer` if possible~

#### Reduce unused CSS

Dashicons, Gutenberg block and component styles are flagged, this is a known issue with using these packages on the frontend.

#### Reduce unused JS

Google tag manager is flagged, which presumably we need

adamwoodnz:
Lighthouse scores have improved. We can reopen for any further work.

PR #237, opened 2023-10-11
Remove unnecessary blocking style requests

adamwoodnz:
See #233 

The [current lighthouse performance score](https://mc.wordpress.org/wporg-site-stats.php?date=2023-10-10) is consistently < 80.

In my sandbox it's between 80 and 90 before these changes. The biggest contributing factor is the LCP metric, and a major factor in that is render blocking styles for dashicons and jetpack:

Screenshot 2023-10-11 at 12 58 41 PM

We don't need either of these on the frontend _as far as I can see_, so this PR fixes that opportunity by removing them. This results in a score > 90, due to an improved LCP metric:

Screenshot 2023-10-11 at 1 00 01 PM

The initial server response time is bad for me accessing my sandbox in NZ, and is considerably better in prod, so I think we can work on these resource loading optimisations first, and look at server response next if necessary.

### Testing

Best to use a sandbox to be as close as possible to prod, and an incognito tab so that no extensions can interfere with results. Run a Lighthouse performance test with and without these changes and compare.

Check that Jetpack and the admin in general still have their required CSS loaded.

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/functions.php
+++ b/source/wp-content/themes/wporg-showcase-2022/functions.php
@@ -17,6 +17,7 @@ require_once __DIR__ . '/inc/block-config.php';
 
 // Filters and Actions
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
+add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\maybe_dequeue_assets' );
 add_action( 'init', __NAMESPACE__ . '\setup_theme' );
 add_action( 'wp_head', __NAMESPACE__ . '\add_social_meta_tags' );
 add_action( 'pre_get_posts', __NAMESPACE__ . '\modify_search_query' );
@@ -39,6 +40,8 @@ add_filter( 'grunion_contact_form_redirect_url', __NAMESPACE__ . '\jetpack_redir
 add_filter( 'grunion_should_send_email', '__return_false' );
 // Enable auto-fill using user information.
 add_filter( 'jetpack_auto_fill_logged_in_user', '__return_true' );
+// Remove Jetpack CSS on frontend
+add_filter( 'jetpack_implode_frontend_css', '__return_false', 99 );
 
 /**
  * Enqueue scripts and styles.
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-showcase-2022/functions.php
```diff
@@ -39,6 +40,8 @@
 add_filter( 'grunion_should_send_email', '__return_false' );
 // Enable auto-fill using user information.
 add_filter( 'jetpack_auto_fill_logged_in_user', '__return_true' );
+// Remove Jetpack CSS on frontend
+add_filter( 'jetpack_implode_frontend_css', '__return_false', 99 );
```
adamwoodnz, line 44: See https://css-tricks.com/snippets/wordpress/removing-jetpack-css/

StevenDufresne: Good find. Looping in @pkevan since he has more experience with JetPack than I do. The front-end and form still appear to work as expected. 👍

StevenDufresne (approved):
This does work as advertised and improved page load. I pinged in the comment but I don't think we need to wait on him for a review.

source/wp-content/themes/wporg-showcase-2022/functions.php
adamwoodnz: Yeah the [Jetpack docs](https://developer.jetpack.com/hooks/jetpack_implode_frontend_css/) describe the intention of this quiet differently, but I've found multiple sources using it for this purpose.

pkevan: When there is something like this found, it's usually best to highlight it with Jetpack themselves, as it could be an opportunity on their side to optimise or a bug in the implementation itself.

renintw (approved):
👍 This goes with the same results as yours on my side.

source/wp-content/themes/wporg-showcase-2022/functions.php
adamwoodnz: > highlight it with Jetpack themselves

Good call, I'll raise this and see if there's a better way to remove this CSS too, as neither `dequeue` or `deregister` work.

PR #238, opened 2023-10-11
Increase local nav spacing

### LINKED ISSUE #236: Increase the space between local nav links
To the space between local nav links

> In the Figma, the gap between submenu items is 24px. I know that's off-grid, I wonder if it's a piece that Francisco missed. We can go with the 20px gap for now and revisit when he's back to see if it's 20 or 30 or indeed 24. But if it's easy, might want to go with 24 for now, and then correct across the componentry at a later time.

Yes. I missed that change, sorry for that. 20px looks too tight next to the global header.

_Originally posted by @fcoveram in https://github.com/WordPress/wporg-showcase-2022/issues/151#issuecomment-1754800949_

adamwoodnz:
Fixes #236 

Changes the spacing between the local nav items from the default block spacing of 20px to 24px.

I've chosen to use local block styles, rather than updating this across all sites using the local-navigation block styles, as [the current local-navigation styles](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss) look like they're lightweight and intended to allow inner blocks to be styled independently.

The downside of this is that the navigation block gap setting doesn't allow independent configuration of column gap and row gap, so the mobile version also gets 24px vertical spacing (see screenshot below). If we controlled this via the local-navigation block styles we could style independently. Is this worth doing @WordPress/meta-design ?

### Screenshots

| Front (_nav-front) | Single (_nav) | Thanks (_nav-dark) |
|-|-|-|
| localhost_8888_(Desktop) | localhost_8888_age-of-union_(Desktop) | localhost_8888_submit-a-wordpress-site_thanks_(Desktop) |

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/patterns/_nav-dark.php
+++ b/source/wp-content/themes/wporg-showcase-2022/patterns/_nav-dark.php
@@ -11,7 +11,7 @@
 <!-- wp:wporg/local-navigation-bar {"backgroundColor":"charcoal-1","style":{"position":{"type":"sticky"},"elements":{"link":{"color":{"text":"var:preset|color|white"},":hover":{"color":{"text":"var:preset|color|white"}}}}},"textColor":"white","fontSize":"small"} -->
 	<!-- wp:site-title {"level":0,"fontSize":"small"} /-->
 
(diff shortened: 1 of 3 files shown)


StevenDufresne (approved):

renintw (approved):
I can see it changed to 24px 👍

adamwoodnz:
Merging this, we can update the mobile styles later if necessary.

PR #227, opened 2023-10-06
Submission: Fix messed up thanks you screen dot pattern

renintw:
Fixes #224

image

#### TODO (Maybe)

- [x] Make the dot pattern in https://wordpress.org/showcase-v2/submit-a-wordpress-site/ (Login/Non-login) also align with the text on both sides.

PR #240, opened 2023-10-11
Fix dots pattern on mobile log in screen & thank you screen

### LINKED ISSUE #226: Mobile: Log in + Thank you screens
Can we update these two screens on mobile.

login-thanks-mobile-screens

I don't think we need to do anything else then expanding the section itself a bit, and move whole paragraph element more below, plus edit rows for copy — as in screenshot. 

Let me know if any help is needed.

renintw:
Fixes #226 

This PR first removes the style from the old design. Then it also ensures the distance between the dot pattern and the text is between 30-50 (submit page) or ~30 & 70~ **40-60** (Login & Thank you screen). 

Additionally, because the Cover Block has an automatically changing margin-top, it makes the CSS control of the distance more complicated on different screens (On smaller screens, the margin-top of the cover disappears, causing the text and dots pattern to overlap.). Plus, I don't see the necessity of the Cover Block here as we only have text at the moment. It might also be a part of the old design. So it has been removed.

[Figma](https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase?type=design&node-id=3526-68710&mode=design&t=8ymL9jC6TvFY4MjX-0)

#### Screenshots

|  | Submit Page (margin bottom: 30px-50px) | Login Page (~30px & 70px~ 40px-60px) | Thank You Page (~30px & 70px~ 40px-60px) |
| -- |-- | -- | -- |
| Bigger Screen | Screen Shot 2023-10-12 at 1 53 37 AM | Screen Shot 2023-10-12 at 1 52 57 AM | Screen Shot 2023-10-12 at 1 53 52 AM |
| Mobile Screen | Screen Shot 2023-10-12 at 1 54 28 AM | Screen Shot 2023-10-12 at 1 54 50 AM | Screen Shot 2023-10-12 at 1 54 12 AM |

> [!NOTE]
> ~Currently, the margin between the dot pattern and text on the Login Page and Thank You Page is:~
>
> ~30px in the mobile view and 70px in non-mobile views (> 599px).~
>
> ~This is according to the Figma design. However, if we want a progressive adjustment like the Submit Page (using clamp to vary the margin between 30-50px), the Login Page and Thank You Page can only vary between 40-60px. This is because after the `--wp--preset--spacing--40`, we only have `--wp--preset--spacing--50: clamp(40px, calc(5vw + 10px), 60px);`.~
>
> ~This means the mobile screen would be at 40px (which is acceptable per the previous discussion), and the maximum would be 60px (This hasn't been discussed yet, might need to see if this is acceptable since the Figma design specifies 70px, a difference of 10px)~

👆 --wp--preset--spacing--50 is used eventually as per the feedback in the comment below.

> [!NOTE]
> Only the Cover Block of the login page has been removed here. The Cover Block for the Thank You Page needs to be edited on the website.

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/patterns/log-in.php
+++ b/source/wp-content/themes/wporg-showcase-2022/patterns/log-in.php
@@ -7,39 +7,32 @@
 
 ?>
 
-<!-- wp:cover {"dimRatio":0,"minHeight":45,"minHeightUnit":"vh","isDark":false,"align":"wide","style":{"spacing":{"padding":{"top":"0","bottom":"0","left":"0","right":"0"}}}} -->
-<div class="wp-block-cover alignwide is-light" style="padding-top:0;padding-right:0;padding-bottom:0;padding-left:0;min-height:45vh">
-	<span aria-hidden="true" class="wp-block-cover__background has-background-dim-0 has-background-dim"></span>
-	<div class="wp-block-cover__inner-container">
-		<!-- wp:group {"align":"wide","style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
-		<div class="wp-block-group alignwide">
-			<!-- wp:heading {"textAlign":"center","level":1,"align":"wide","style":{"spacing":{"margin":{"bottom":"var:preset|spacing|30"}}},"fontSize":"heading-3"} -->
-			<h1 class="wp-block-heading alignwide has-text-align-center has-heading-3-font-size" style="margin-bottom:var(--wp--preset--spacing--30)"><?php esc_html_e( 'Log in to submit a site', 'wporg' ); ?></h1>
-			<!-- /wp:heading -->
+<!-- wp:group {"align":"wide","style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignwide">
+	<!-- wp:heading {"textAlign":"center","level":1,"align":"wide","style":{"spacing":{"margin":{"bottom":"var:preset|spacing|30"}}},"fontSize":"heading-3"} -->
+	<h1 class="wp-block-heading alignwide has-text-align-center has-heading-3-font-size" style="margin-bottom:var(--wp--preset--spacing--30)"><?php esc_html_e( 'Log in to submit a site', 'wporg' ); ?></h1>
+	<!-- /wp:heading -->
 
-			<!-- wp:paragraph {"align":"center"} -->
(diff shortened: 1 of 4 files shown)


renintw:
@jasmussen @marko-srb 

Based on the first Note in the description above, if we change the margin between the dot pattern and text to 40px on mobile (currently at 30px), and the maximum margin on non-mobile to 60px (70px as per Figma), would that be ok?

marko-srb:
@renintw 

> Based on the first Note in the description above, if we change the margin between the dot pattern and text to 40px on mobile (currently at 30px), and the maximum margin on non-mobile to 60px (currently 70px as per Figma), would that be ok?

That works well, just tested in design. Feel free to proceed.

renintw:
Thanks @marko-srb.

@adamwoodnz could you take a look at the PR? The question asked before has been clarified and I've updated the commit message and the results in the PR description. Thanks 🙏

adamwoodnz (commented):
Works for me. I notice that the dots are aligned left, so that more dots are added from the right as the screen gets wider. On small screens this looks a little unbalanced IMO

Have we considered positioning the dots centrally, possibly only on smaller screens where it's more noticeable?

| background-position: left | background-position: centre |
|-|-|
| localhost_8888_submit-a-wordpress-site_thanks_(Samsung Galaxy S20 Ultra) (1) | localhost_8888_submit-a-wordpress-site_thanks_(Samsung Galaxy S20 Ultra) (2) |

renintw:
yeah, that change is from this [PR](https://github.com/WordPress/wporg-showcase-2022/pull/227) and there are some discussions about this.

> Have we considered positioning the dots centrally, possibly only on smaller screens where it's more noticeable?

@jasmussen do you think we should position the dots centrally for all sizes of screen or just smaller screen in this case?

jasmussen:
Good eyes.

Most important thing: whatever we decide, we should do that everywhere. I believe that Marko designed it to be centered initially, though I suggested we should left-align because then the dots always line up with News on the left. But I do see that when we reach the mobile breakpoint, it's less elegant with the uneven padding left and right. 

I'm happy to go full centered, so long as we do it consistently, and if Marko chimes in otherwise, I'll defer to him!

adamwoodnz:
Also I notice we've got widows on mobile

Screenshot 2023-10-18 at 8 31 02 AM

adamwoodnz (approved):
Checked at sizes of iPhone 12, Samsung Galaxy S20, iPad and desktop, and dots LGTM

renintw:
`white-space: nowrap`  would make it overflow, I've opened an issue to address this.

PR #241, opened 2023-10-11
Remove Flavors UI

### LINKED ISSUE #230: Consider merging "Flavors" with "Tags"
Flavors were implemented in [Showcase v2](https://wordpress.org/showcase-v2/) to keep the functionality consistent with the [current Showcase](https://wordpress.org/showcase/). After using the v2 site, the third taxonomy feels burdensome as a user. There are too many things to search by, and it is unclear what "Flavor" means. 

| Homepage | Single Site |
| --- | --- |
| image | image |

Therefore, I propose that we convert the existing Flavors into tags. We can ignore "WordPress.org", so only "WordPress Multisite" and "WordPress VIP" would be added as tags. Then:

- Update any sites with these Flavors with the new corresponding tags
- Remove the Flavor filter from the Homepage, Search/Filter Results, and All Sites pages
- Remove the Flavor row from the Single Sites template
- Remove the Flavor selector from the Submit a Site form

I don't think we should remove the Flavor taxonomy entirely at the moment, just remove the visual presence on the front end. There may be a use case in the future. But for now, tags and categories suffice. 

cc @marko-srb @jasmussen @eidolonnight

outdoor2kode:
Closes #230 

This PR removes Flavors from Homepage Filter, Search Page Filter, Archive Page Filter, Browse Page, and Single Page.

|  | Screenshots |
| -- | -- |
| Homepage | image |
| Serach Page | image | 
| Archive Page | image | 
| Browse Page | image |
|  Single Page | image |

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/patterns/_site-grid-featured.php
+++ b/source/wp-content/themes/wporg-showcase-2022/patterns/_site-grid-featured.php
@@ -24,7 +24,6 @@
 		<div class="wp-block-group wporg-query-filters">
 			<!-- wp:wporg/query-filter {"key":"post_tag"} /-->
 			<!-- wp:wporg/query-filter {"key":"category"} /-->
-			<!-- wp:wporg/query-filter {"key":"flavor"} /-->
 		</div>
 		<!-- /wp:group -->
 	</div>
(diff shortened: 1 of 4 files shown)


ndiego:
After further discussion, let's put this PR on hold. Sorry for the extra work here! For now, we will keep feature parity with the current Showcase (which includes Flavors) and then we will reassess in the future.

PR #239, opened 2023-10-11
Improve screen reader experience on site meta lists

adamwoodnz:
Fixes #235 

The flex layout on the site meta list items causes the nested elements to be read individually, rather than as a complete `label: value` pair.

This PR adds an `aria-label` attribute on the list items which is read as intended, eg. 'Category: Creative'. Nested elements have `aria-hidden` unless they contain a URL, as this still needs to be accessible. This reduces duplicated content unless necessary.

On the home page the labels are visibly hidden, so there are changes to not render them at all in this case, and rely on the `aria-label` instead.

There is no visual change.

The resulting HTML looks like this:

#### Home 

```html
<ul>
	<li class="is-meta-post_title" aria-label="Site title: Age of Union">
		<span><a href="http://localhost:8888/age-of-union/">Age of Union</a></span>
	</li>
	<li class="is-meta-domain" aria-label="URL: ageofunion.com">
		<span><a class="external-link" href="https://ageofunion.com/" target="_blank" rel="noopener noreferrer">ageofunion.com</a></span>
	</li>
	<li class="is-meta-category" aria-label="Category: Community">
		<span><a href="http://localhost:8888/category/community/" rel="tag">Community</a></span>
	</li>
</ul>
```

#### Single

```html
<ul>
	<li class="is-meta-country" aria-label="Country: Canada">
		<strong aria-hidden="true">Country</strong>
		<span aria-hidden="true">Canada</span>
	</li>
	<li class="is-meta-category" aria-label="Category: Community">
		<strong aria-hidden="true">Category</strong>
		<span><a href="http://localhost:8888/category/community/" rel="tag">Community</a></span>
	</li>
	<li class="is-meta-published" aria-label="Published: September 2023">
		<strong aria-hidden="true">Published</strong>
		<span aria-hidden="true">September 2023</span>
	</li>
	<li class="is-meta-post_tag" aria-label="Site tags: Conservation, Environment, Nonprofit">
		<strong aria-hidden="true">Site tags</strong>
		<span><a href="http://localhost:8888/tag/conservation/" rel="tag">Conservation</a>, <a href="http://localhost:8888/tag/environment/" rel="tag">Environment</a>, <a href="http://localhost:8888/tag/nonprofit/" rel="tag">Nonprofit</a></span>
	</li>
</ul>
```

## Testing

Use a screen reader to navigate the home and single site pages, checking how the screen reader interprets the meta list.

[...]

PR #245, opened 2023-10-12
Improve screen reader experience for site meta (v2)

### LINKED ISSUE #235: Accessibility improvements for the Site metadata table
Copied from WordPress Slack: https://wordpress.slack.com/archives/C04U953K77A/p1696899077421269?thread_ts=1696525287.393319&cid=C04U953K77A

> Really need to do something about the lists. It would be nice if it was read as Category: X but instead the list item is split. I wonder if the usage of flex is causing this?

On screenreaders, the content in the Site metadata table is not being read well. What could we do to improve this?

image

adamwoodnz:
Fixes #235 
Alternative to #239 

The PR uses separate content for screen reader and visual layout, so that Voiceover and NVDA will both read the entire content of the list item.

#### Home (no visible label)

```html
<ul>
	<li class="is-meta-post_title">
		<span class="screen-reader-text">Site title: <a href="http://localhost:8888/google-ventures/">GV (Google Ventures)</a></span>
		<span aria-hidden="true"><a href="http://localhost:8888/google-ventures/">GV (Google Ventures)</a></span>
	</li>
	<li class="is-meta-domain">
		<span class="screen-reader-text">URL: <a class="external-link" href="https://www.gv.com/" target="_blank" rel="noopener noreferrer">gv.com</a></span>
		<span aria-hidden="true"><a class="external-link" href="https://www.gv.com/" target="_blank" rel="noopener noreferrer">gv.com</a></span>
	</li>
	<li class="is-meta-category">
		<span class="screen-reader-text">Category: <a href="http://localhost:8888/category/business/" rel="tag">Business</a></span>
		<span aria-hidden="true"><a href="http://localhost:8888/category/business/" rel="tag">Business</a></span>
	</li>
</ul>
```

#### Single

```html
<ul>
	<li class="is-meta-country">
		<span class="screen-reader-text">Country: United States</span>
		<span aria-hidden="true"><strong>Country</strong><span>United States</span></span>
	</li>
	<li class="is-meta-category">
		<span class="screen-reader-text">Category: <a href="http://localhost:8888/category/business/" rel="tag">Business</a></span>
		<span aria-hidden="true"><strong>Category</strong><span><a href="http://localhost:8888/category/business/" rel="tag">Business</a></span></span>
	</li>
	<li class="is-meta-published">
		<span class="screen-reader-text">Published: September 2023</span>
		<span aria-hidden="true"><strong>Published</strong><span>September 2023</span></span>
	</li>
	<li class="is-meta-post_tag">
		<span class="screen-reader-text">Site tags: <a href="http://localhost:8888/tag/entrepreneur/" rel="tag">Entrepreneur</a>, <a href="http://localhost:8888/tag/startups/" rel="tag">Startups</a>, <a href="http://localhost:8888/tag/technology/" rel="tag">Technology</a></span>
		<span aria-hidden="true"><strong>Site tags</strong><span><a href="http://localhost:8888/tag/entrepreneur/" rel="tag">Entrepreneur</a>, <a href="http://localhost:8888/tag/startups/" rel="tag">Startups</a>, <a href="http://localhost:8888/tag/technology/" rel="tag">Technology</a></span></span>
	</li>
</ul>
```

Using specific screen reader content means we can avoid nested elements or visual space between elements, which cause screen readers to not read the list item as one.

Using NVDA where there are specific keys for navigating lists (L) and list items (I), the whole first item is read out when either key is used, eg. 'Country: United States' or 'Category: Creative (link)'. Full transcript from focusing list, then navigating items:

> main landmark    list  with 4 items  Country: Germany
Category:   Creative
Published: September 2023
Site tags:   Architecture  link    ,   Development  link    ,   Sustainability

Using Voiceover the nested anchor tags still require another keyboard navigation with right arrow, and are read individually, but the country and published items (without links) are read as one.

My understanding is that JAWS and NVDA cover the vast majority of screen reader users, so I think optimising for them is the best choice.

There are no visual changes.

### Testing

Use a screen reader to navigate the home and single site pages, checking how the screen reader interprets the meta list.

It should announce the list, then read each list item as 'label: value', and you should be able to navigate to the link within if appropriate (only country and published don't have links).

The label text should not be read twice.

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-meta-list/index.php
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-meta-list/index.php
@@ -99,12 +99,19 @@ function render( $attributes, $content, $block ) {
 		$value = get_value( $field['type'], $field['key'], $block->context['postId'] );
 
 		if ( ! empty( $value ) ) {
+			$value = wp_kses_post( $value );
+
 			$list_items[] = sprintf(
-				'<li class="is-meta-%1$s"><strong%2$s>%3$s</strong> <span>%4$s</span></li>',
+				'<li class="is-meta-%1$s">
+					<span class="screen-reader-text">%2$s: %3$s</span>
+					<span aria-hidden="true">%4$s</span>
+				</li>',
 				$field['key'],
-				$show_label ? '' : ' class="screen-reader-text"',
 				$field['label'],
-				wp_kses_post( $value )
+				$value,
+				$show_label
+					? sprintf( '<strong>%1$s</strong><span>%2$s</span>', $field['label'], $value )
+					: $value,
 			);
 		}
 	}
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-meta-list/style.scss
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-meta-list/style.scss
@@ -10,8 +10,6 @@
 	}
 
 	li {
-		display: flex;
-		gap: var(--wp--preset--spacing--10);
 		padding: var(--wp--preset--spacing--10) var(--wp--preset--spacing--20);
 		border-bottom: 1px solid;
 		border-color: inherit;
@@ -20,24 +18,29 @@
 			border-bottom: none;
 		}
 
-		strong {
-			flex-shrink: 0;
-			font-weight: 400;
-		}
+		span[aria-hidden="true"] {
+			display: flex;
+			gap: var(--wp--preset--spacing--10);
 
-		span {
-			margin-inline-start: auto;
-			text-align: right;
-			color: var(--wp--preset--color--charcoal-3);
-		}
-
-		@media (max-width: 380px) {
-			flex-direction: column;
-			gap: 0;
+			strong {
+				flex-shrink: 0;
+				font-weight: 400;
+			}
 
 			span {
-				margin-inline-start: unset;
-				text-align: initial;
+				margin-inline-start: auto;
+				text-align: right;
+				color: var(--wp--preset--color--charcoal-3);
+			}
+
+			@media (max-width: 380px) {
+				flex-direction: column;
+				gap: 0;
+
+				span {
+					margin-inline-start: unset;
+					text-align: initial;
+				}
 			}
 		}
 	}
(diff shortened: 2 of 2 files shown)


adamwoodnz:
The last version was no good using NVDA @alexstine. I'll merge this one for testing instead 🤞

adamwoodnz:
This solution has dropped our overall Lighthouse accessibility score due to focusable elements being `aria-hidden` 😞 

If it improves the experience with these lists, maybe that doesn't matter too much. [I didn't have any success using different layout techniques](https://github.com/WordPress/wporg-showcase-2022/issues/235#issuecomment-1758688681) and keeping the label and value read together. Any space between them causes the screen reader to separate them.

adamwoodnz:
Hey @alexstine this is merged, would you be able to give it a spin on []([staging](https://wordpress.org/showcase-v2/)) please? 

An alternative approach which works ok in NVDA and doesn't harm the Lighthouse score, is to wrap the previous contents of the list item in `<div role="group">`, eg.

```html
<li class="is-meta-post_tag">
	<div role="group">
		<strong>Site tags</strong> <span><a href="http://localhost:8888/tag/entrepreneur/" rel="tag">Entrepreneur</a>, <a href="http://localhost:8888/tag/startups/" rel="tag">Startups</a>, <a href="http://localhost:8888/tag/technology/" rel="tag">Technology</a></span>
	</div>
</li>
```

This does result in the list items being read as one (using I in NVDA), however not when the list is initially focused (with L). The transcript when focusing the list and then navigating the items is:

> main landmark    list  with 4 items  Country
Category  Publication
Published  September 2023
Site tags  Journalism  link    ,   Media  link    ,   Newspaper

I notice the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role#description) suggest restrictions on the usage of this role within lists:

> When a group is used in the context of list, limit the children of the group to listitem elements. In this case, it is recommended to use multiple ordered or unordered lists, <ol> or <ul>, with nested <li> children.

The technique also doesn't really help in Voiceover, but I don't think that is so important.

Thoughts?

alexstine:
@adamwoodnz So, you cannot add `aria-hidden="true"` without also adding `tabindex="-1"` which would also take it out of tab order for sighted users. Using `role="group"` is really not valid here. Is there a way we can avoid using flex box on the list items?

Thanks.

adamwoodnz:
Thanks for checking

> you cannot add aria-hidden="true" without also adding tabindex="-1" which would also take it out of tab order for sighted users

Ah I got so busy navigating using special keys I forgot about standard tab behaviour 🤦 

> Is there a way we can avoid using flex box on the list items?

So far I have not been able to find a way to make this spread out design work with screen readers. I've tried `flex`, `grid` and `table`. I'll try absolute positioning, but that gets brittle.

PR #246, opened 2023-10-12
Stop loading `wp-mediaelement` styles on the frontend

adamwoodnz:
These styles also seem to have no use on the frontend, and are also being flagged under render blocking resources.


StevenDufresne (approved):
Yep. That fine. We can revert later if needed.

renintw (approved):
👍

PR #248, opened 2023-10-12
Query Filters: Replace overflow-y with overflow-x

renintw:
Related: https://github.com/WordPress/wporg-mu-plugins/pull/473

In showcase, the "Query Filters" only overflow horizontally ([noWrap is set](https://github.com/WordPress/wporg-showcase-2022/blob/main/source/wp-content/themes/wporg-showcase-2022/patterns/_site-grid-featured.php#L23)), so overflow-x is used in place of overflow-y.


adamwoodnz (approved):
👍 Works and makes sense to me

Issue #234, opened 2023-10-10
Showcase v2 Feedback


ndiego:
Copied over from WP Slack: https://wordpress.slack.com/archives/C04U953K77A/p1696925380920099?thread_ts=1696525287.393319&cid=C04U953K77A

- As a user, I don't really understand what the links on the left of the 'homepage' banner are/mean. See attachment.
  - What is "Meta Newsroom"? Why is there a Facebook link? Why would I click 'Business' / where is this taking me? We're assuming context that the user doesn't have. Can we rework these into a simple paragraph?
- Canonical URL on search pages uses the 'pretty' URL format; it should use the ?s= format.
- Archive pages (including search results and tags) don't have a H1, so it's not clear what I'm looking at (nb, this should also consider the paginated state).
- Archive pages assume that the user knows where they are / what they're looking at. https://wordpress.org/showcase-v2/browse/ has no context/information about what I'm browsing / what these filters are. Descriptive + helpful introductory on all pages/templates pls :smile:
- Can we kill/hide all tags with fewer than 3(?) members? E.g., https://wordpress.org/showcase-v2/tag/space/ isn't useful :laughing:
- Page <title>s are inconsistent about whether they include (or end in)  "WordPress Showcase"; can we define and use a consistent pattern throughout which includes this?
- I'm also not clear on the difference between categories and tags (good old WordPress, eh? :wink:)
- In the breadcrumb, we have 'Home' immediately below 'Showcase'; can we remove 'Home' (as it doesn't link to the homepage), and have Home > Showcase > etc? Or if that's contentious, just replace 'Home' with 'Showcase' in the breadcrumbs.
- The Next/Previous nav on individual items is confusing, and a worse experience than the 'Related sites' nav. Can we remove it?
- Is there a reason that we're setting noreferrer on links to sites?

ndiego:
cc @jasmussen @marko-srb

jasmussen:
Good stuff here. Marko is a bit busy at the moment so I'll respond to the best of my ability. In general, agree with most of this, so the stuff I didn't comment on/add context to below, would be good to just run with.

> What is "Meta Newsroom"? Why is there a Facebook link? Why would I click 'Business' / where is this taking me? We're assuming context that the user doesn't have. Can we rework these into a simple paragraph?

Conceptually it's the feature version of this box;

[...]

PR #251, opened 2023-10-16
Remove `noreferrer` from Site Link and Site Meta List blocks.

### LINKED ISSUE #250: Remove `noreferrer` from links in Site Link block
Per the feedback in #234, there really shouldn't be `noreferrer` on the site links in the Showcase. It's probably a good thing for receiving sites to see wp.org referrals.

The `noreferrer` needs to be removed here: https://github.com/WordPress/wporg-showcase-2022/blob/94d746f4d7895528ba3ed14fb4f9b59b3b86be78/source/wp-content/themes/wporg-showcase-2022/src/site-link/index.php#L42

ndiego:
Fixes #250 

Removed `noreferrer` from the link in the Site Link and Site Meta List blocks per feedback in #234

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-link/index.php
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-link/index.php
@@ -39,7 +39,7 @@ function render( $attributes, $content, $block ) {
 
 	$wrapper_attributes = get_block_wrapper_attributes( array( 'class' => 'external-link' ) );
 	return sprintf(
-		'<p %1$s><a href="%2$s" target="_blank" rel="noopener noreferrer">%3$s</a></p>',
+		'<p %1$s><a href="%2$s" target="_blank" rel="noopener">%3$s</a></p>',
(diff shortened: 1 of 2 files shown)


renintw:
I guess the original decision was based on a practice suggesting that using "noopener noreferrer" together is a safe approach. That's probably why it was implemented as such. It seems like "noreferrer" could be safely removed here in the showcase, as I haven't seen URLs containing sensitive user information.

adamwoodnz (commented):
LGTM, seems like we'd want to remove it here too?

https://github.com/WordPress/wporg-showcase-2022/blob/main/source/wp-content/themes/wporg-showcase-2022/src/site-meta-list/index.php#L143

ndiego:
> LGTM, seems like we'd want to remove it in the site meta list URL entry too?

Ah, good catch. Just updated the PR to include that.

adamwoodnz (approved):
LGTM :shipit:

PR #268, opened 2024-01-24
Site Screenshot block: Update to support latest Interactivity API

### LINKED ISSUE #267: Site Screenshots break with GB 17.2+ (interactivity API updates)
See https://github.com/WordPress/wporg-mu-plugins/issues/490. The site screenshot block uses the interactivity API to load images, and the API structure has changed, causing JS errors on the page.

This is not currently a problem on production, because we're stuck on GB 16.8.

ryelle:
**Merge alongside https://github.com/WordPress/wporg-mu-plugins/pull/561, so that production has current Gutenberg version.**

Update the dependencies so that we can use the `viewModule` block.json entry, which wp-scripts will correctly build as a module (required by the interactivity API script).

The updated wp-scripts requires a more modern node version, so I've also updated the version for this project to 20 (see https://github.com/WordPress/wporg-parent-2021/issues/122, we'll want to do this across all projects). I've pushed the package and node version updates to `main`, so this PR only contains the Interactivity API updates.

The changes for the Interactivity API are mostly following [the changelog recommendations](https://github.com/WordPress/gutenberg/discussions/52906#discussioncomment-7810998). The exception is the change to how the screenshots are rendered. Rather than directly manipulating the DOM, the new method uses a combination of directives and state values to show/hide the image once it's loaded. This is more in line with expected behavior from the Interactivity API.

Fixes #267.

**To test**

1. Update composer packages so you have the latest Gutenberg: `composer update`
2. Update node: `nvm install`
3. Update yarn deps: `yarn`
4. Build the theme: `yarn build:theme`
5. Start the env: `yarn wp-env start`
6. Click through the site, all the screenshots should show up correctly— if no local image is attached, there should be a loading spinner, and then load in the mshots image.

### DIFF
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/view.js
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/view.js
@@ -20,7 +20,7 @@ const { actions, state } = store( 'wporg/showcase/screenshot', {
 		},
 		get base64Image() {
 			return getContext().base64Image;
-		}
+		},
 	},
 	actions: {
 		increaseAttempts() {
--- a/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/view.js
+++ b/source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/view.js
@@ -2,7 +2,7 @@
 /**
  * WordPress dependencies
  */
-import { store as wpStore } from '@wordpress/interactivity';
+import { getContext, getElement, store } from '@wordpress/interactivity';
 
 /**
  * Module constants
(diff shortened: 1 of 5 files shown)


source/wp-content/themes/wporg-showcase-2022/src/site-screenshot/view.js
```diff
@@ -2,142 +2,115 @@
...
+					async ( _context ) => {
+						const { attempts, base64Image: _base64Image, shouldRetry } = _context;
+						if ( shouldRetry ) {
+							await actions.fetchImage( src );
+						}
+						if ( attempts >= MAX_ATTEMPTS ) {
+							clearInterval( intervalId );
+							if ( ! _base64Image ) {
+								actions.setHasError( true );
+							}
```
ryelle, line 82: These actions aren't working, because the callback happens outside the scope of context. Same issue happened in `fetchImage` until I refactored it to use the Promise.

Screenshot 2024-01-25 at 6 15 59 PM

adamwoodnz (approved):
Works for me, and changes LGTM :shipit:

